PR #3, opened 2015-07-25
prefix category methods

### LINKED ISSUE #1: Category methods should be prefixed
In order to avoid possible name collisions, category methods on Objective-C objects should be prefixed: [https://developer.apple.com/library/ios/documentation/Cocoa/Conceptual/ProgrammingWithObjectiveC/CustomizingExistingClasses/CustomizingExistingClasses.html](source) via [http://nshipster.com/namespacing/](NSHipster). Instead of:

```objc
[imageView setUpdateWithProgress:YES];
```

it should be:

```objc
[imageView pin_setUpdateWithProgress:YES];
```

ZevEisenberg:
Closes #1 

I don't know much about Appledoc, so you'll probably want to confirm that I generated the docs correctly.


garrettmoon:
Thanks so much for doing this! I'll take a look at getting it merged in on Monday :)

ZevEisenberg:
Resolved merge conflicts. The `testEarlyReturn` is currently failing on my machine, but I'm not sure if I caused it. See #5

jparise:
@ZevEisenberg tests appear to be running fine here (at master).

garrettmoon:
Yup, looking into Travis CI today. I'll also take a look at that test to make sure there aren't race conditions that could be causing it to fail for you locally and not us :)

ZevEisenberg:
Definitely worth checking. I was on a slow connection when I was working on it before, so that could have affected the test.

garrettmoon:
Hmm, I think I pushed up some stuff that created conflicts, I'll look into resolving them when I get to testing this out :)

ZevEisenberg:
Just need to run `pod install`. I'll take care of it.

garrettmoon:
You'll need to rebase against master to get the Travis build to work :D

ZevEisenberg:
I did the thing :zap:

ZevEisenberg:
You might want to prefix the methods in the blog post, as well. I would, but I can't submit a PR.

garrettmoon:
Ha! Yes :)

Pod/Classes/Image Categories/FLAnimatedImageView+PINRemoteImage.m
```diff
@@ -10,92 +10,92 @@
...
 }
 
-- (BOOL)updateWithProgress
+- (BOOL)pin_updateWithProgress
 {
     return [PINRemoteImageCategoryManager updateWithProgressOnView:self];
 }
 
-- (void)setUpdateWithProgress:(BOOL)updateWithProgress
+- (void)setPin_updateWithProgress:(BOOL)updateWithProgress
```
garrettmoon: Shouldn't this be pin_setUpdateWithProgress?

garrettmoon:
Thanks again for this, if you can address the above comment I'll click the merge button :)

Pod/Classes/Image Categories/FLAnimatedImageView+PINRemoteImage.m
ZevEisenberg: Not if you want dot syntax to work.

garrettmoon: That's what I was wondering, but then I looked at the others and they didn't match the pattern. Now I see this is the only one declared as a property. I swear I wrote this thing ;)

garrettmoon:
Thanks so much again!

PR #7, opened 2015-07-26
fix(imports): change to module imports where relevant

jaredsinclair:


jparise:
Thanks, @jaredsinclair!

PR #11, opened 2015-07-30
Fixing the early return test.

garrettmoon:
I forgot about the optimization to execute callbacks before images are stored in the cache.

### DIFF
--- a/Example/PINRemoteImage Tests/PINRemoteImage_Tests.m
+++ b/Example/PINRemoteImage Tests/PINRemoteImage_Tests.m
@@ -315,7 +315,10 @@
     dispatch_semaphore_wait(semaphore, [self timeout]);
     // callback can occur *before* image is stored in cache this is an optimization to avoid waiting on the cache to write.
     // So, wait until it's actually in the cache.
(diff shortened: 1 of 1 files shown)


Example/PINRemoteImage Tests/PINRemoteImage_Tests.m
```diff
@@ -311,7 +311,13 @@ - (void)testEarlyReturn {
     [self.imageManager downloadImageWithURL:[self JPEGURL] completion:^(PINRemoteImageManagerResult *result) {
         dispatch_semaphore_signal(semaphore);
     }];
+    
     dispatch_semaphore_wait(semaphore, [self timeout]);
+    // callback can occur *before* image is stored in cache this is an optimization to avoid waiting on the cache to write.
+    // So, wait until it's actually in the cache.
+    while ([[self.imageManager cache] objectForKey:[self.imageManager cacheKeyForURL:[self JPEGURL] processorKey:nil]] == nil) {
```
jparise: Better to do this with a `for` loop so we can timeout after a predetermined number of attempts.

jparise:
:+1:

PR #12, opened 2015-07-30
Fix QOS Test (and add cleanup)

garrettmoon:
- Remove lower quality images from cache if a higher quality one is to be downloaded
- Document weirdness around QOS and caching
- Fix QOS test

### DIFF
--- a/Example/PINRemoteImage Tests/PINRemoteImage_Tests.m
+++ b/Example/PINRemoteImage Tests/PINRemoteImage_Tests.m
@@ -206,8 +206,9 @@
 
 - (void)waitForImageWithURLToBeCached:(NSURL *)URL
 {
+    NSString *key = [self.imageManager cacheKeyForURL:URL processorKey:nil];
     for (NSUInteger idx = 0; idx < 100; idx++) {
-        if ([[self.imageManager cache] objectForKey:[self.imageManager cacheKeyForURL:URL processorKey:nil]] != nil) {
+        if ([[self.imageManager cache] objectForKey:key] != nil) {
             break;
         }
         sleep(50);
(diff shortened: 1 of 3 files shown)


Example/PINRemoteImage Tests/PINRemoteImage_Tests.m
```diff
@@ -204,6 +204,16 @@ - (void)drawImage:(UIImage *)image
     UIGraphicsEndImageContext();
 }
 
+- (void)waitForImageWithURLToBeCached:(NSURL *)URL
+{
+    for (NSUInteger idx = 0; idx < 100; idx++) {
+        if ([[self.imageManager cache] objectForKey:[self.imageManager cacheKeyForURL:URL processorKey:nil]] != nil) {
```
jparise: You could call `-cacheForKey:processorKey:` outside of the loop.

Example/PINRemoteImage Tests/PINRemoteImage_Tests.m
```diff
@@ -594,6 +605,15 @@ - (void)testQOS
      }];
     dispatch_semaphore_wait(semaphore, [self timeout]);
     
+    //small image should have been removed from cache
+    for (NSUInteger idx = 0; idx < 100; idx++) {
```
jparise: Can you use `-waitForImageWithURLToBeCached` here?

Pod/Classes/PINRemoteImageManager.m
```diff
@@ -1165,23 +1176,35 @@ - (NSUUID *)downloadImageWithURLs:(NSArray *)urls
...
                             processorKey:nil
                                processor:nil
                                 progress:progress
-                              completion:completion
+                              completion:^(PINRemoteImageManagerResult *result) {
+                                  typeof(self) strongSelf = weakSelf;
+                                  //clean out any lower quality images from the cache
+                                  for (NSInteger idx = downloadIdx - 1; idx >= 0; idx--) {
+                                      [[strongSelf cache] removeObjectForKey:[strongSelf cacheKeyForURL:[urls objectAtIndex:idx] processorKey:nil]];
+                                  }
```
jparise: This looks fine but makes me think we should consider a `-removeObjectsForKeys:(NSArray *)` method in PINCache that can do this work in a single locked batch operation.

garrettmoon: That's a good idea.

Example/PINRemoteImage Tests/PINRemoteImage_Tests.m
garrettmoon: No, this is checking for the opposite, that the image is *not* in the cache.

PR #15, opened 2015-07-31
Pass options directly to our early return routine.

jparise:
This is a more flexible approach that will allow this routine to consider the
full set of options when evaluating an early return.

Also, the `queue` parameter was never being used so it has been removed.


garrettmoon:
Looks good to me.

PR #16, opened 2015-08-01
Fix concurrency issue: Only alter completions dictionary while locked

jmah:


garrettmoon:
Looks good to me, thanks for catching this!

PR #20, opened 2015-08-01
Set QOS / Thread priority

garrettmoon:
Surprisingly, it seems that leaving the qos priority to default does not actually lead to a qos of background like the documentation states. I've found that the QOS gets upgraded to user interactive in some cases. This explicitly sets the QOS on all operations to background so we don't affect scroll performance.

### DIFF
--- a/Pod/Classes/PINRemoteImageManager.m
+++ b/Pod/Classes/PINRemoteImageManager.m
@@ -1244,9 +1244,8 @@ typedef void (^PINRemoteImageManagerDataCompletion)(NSData *data, NSError *error
 {
     NSBlockOperation *operation = [NSBlockOperation blockOperationWithBlock:block];
     operation.queuePriority = operationPriorityWithImageManagerPriority(priority);
-    operation.qualityOfService = NSOperationQualityOfServiceBackground;
 #if defined(__IPHONE_8_0)
-    operation.qualityOfService = NSQualityOfServiceBackground;
+    operation.qualityOfService = NSOperationQualityOfServiceBackground;
 #else
     operation.threadPriority = 0.2;
 #endif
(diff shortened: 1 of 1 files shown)


Pod/Classes/PINRemoteImageManager.m
```diff
@@ -1241,6 +1244,12 @@ - (void)pin_addOperationWithQueuePriority:(PINRemoteImageManagerPriority)priorit
 {
     NSBlockOperation *operation = [NSBlockOperation blockOperationWithBlock:block];
     operation.queuePriority = operationPriorityWithImageManagerPriority(priority);
+    operation.qualityOfService = NSOperationQualityOfServiceBackground;
+#if defined(__IPHONE_8_0)
+    operation.qualityOfService = NSQualityOfServiceBackground;
```
jparise: It looks like this sets `operation.qualityOfService` twice.

garrettmoon: Good catch, thank you.

Pod/Classes/PINRemoteImageManager.m
```diff
@@ -1241,6 +1244,11 @@ - (void)pin_addOperationWithQueuePriority:(PINRemoteImageManagerPriority)priorit
 {
     NSBlockOperation *operation = [NSBlockOperation blockOperationWithBlock:block];
     operation.queuePriority = operationPriorityWithImageManagerPriority(priority);
+#if defined(__IPHONE_8_0)
+    operation.qualityOfService = NSOperationQualityOfServiceBackground;
+#else
+    operation.threadPriority = 0.2;
```
jparise: So we only set the thread priority for iOS 7, correct?  If we assigned a thread priority unconditionally, would that conflict with our QoS preference on iOS 8+?

garrettmoon:
threadPriority was deprecated in iOS 8

PR #24, opened 2015-08-03
Release our lock when we encounter a canceled task

jparise:
This early-out condition would exit while still holding our lock, resulting in
Very Bad Things (tm) down the line.


garrettmoon:
:thumbsup:

PR #26, opened 2015-08-04
Fix under os8

garrettmoon:
Apparently I'd forgotten how to check os at runtime. Sigh.


garrettmoon:
Tested on iOS 7 & iOS 8

jparise:
:+1:

PR #27, opened 2015-08-04
Fixes podspec syntax for git tag.

fonkadelic:
The podspec contains a syntax error for the git tag value, which prevents Cocoapods from installing versions greater than `1.0.0`.

You can run `pod spec lint` in the project directory to validate the podspec.


fonkadelic:
Ha... just saw that is fixed now. Cool!

garrettmoon:
Yes! Just fixed it now :)

The linter actually didn't catch it which is why it took so long to find :)

Issue #39, opened 2015-08-14
Decoded images ignore image orientation


waltflanagan:
Using UIImageView `-pin_setImageFromURL:` and related methods lose orientation metadata and display incorrectly.  

This seems to be because that are decoded using `pin_decodedImageWithCGImageRef:` don't take the images orientation into account when drawing the CGImageRef into the context. 

Repro steps:
```swift 
imageView.pin_setImageFromURL(NSURL(string: "https://www.dropbox.com/s/nakmlzyqkqd8ie8/04cd607b-d5ba-4f4a-9344-77c8ef94c291--2015-08-12-08-41-27.jpg?dl=1"))
```

That image was taken with an iPhone 6, portrait orientation.  It appears rotated when set on the imageview via `pin_setImageFromURL`.  Viewing it in a browser, preview, etc show it correctly.

garrettmoon:
Awesome, thank you for the pull request!

garrettmoon:
Fixed by #40

PR #40, opened 2015-08-14
Addresses #39

waltflanagan:
Converts the embeded EXIF orientation into UIImageOrientation to be used when creating the resulting UIImage

### DIFF
--- a/Pod/Classes/Categories/UIImage+DecodedImage.m
+++ b/Pod/Classes/Categories/UIImage+DecodedImage.m
@@ -51,7 +51,7 @@
         CGImageRef imageRef = CGImageSourceCreateImageAtIndex(imageSourceRef, 0, (CFDictionaryRef)@{(NSString *)kCGImageSourceShouldCache : (NSNumber *)kCFBooleanFalse});
         if (imageRef) {
             
-            UIImageOrientation orientation = [self pin_UIImageOrienationFromImageSource:imageSourceRef];
+            UIImageOrientation orientation = pin_UIImageOrienationFromImageSource(imageSourceRef);
             
             decodedImage = [self pin_decodedImageWithCGImageRef:imageRef orientation:orientation];
             
@@ -108,8 +108,8 @@
     return decodedImage;
 }
 
-+(UIImageOrientation) pin_UIImageOrienationFromImageSource:(CGImageSourceRef)imageSourceRef {
-    
+
+UIImageOrientation pin_UIImageOrienationFromImageSource(CGImageSourceRef imageSourceRef) {
     UIImageOrientation orientation = UIImageOrientationUp;
     
     if (imageSourceRef != nil) {
@@ -160,8 +160,8 @@
     }
     
     return orientation;
-    
 }
+
 
 @end
 
--- a/Pod/Classes/Categories/UIImage+DecodedImage.m
+++ b/Pod/Classes/Categories/UIImage+DecodedImage.m
@@ -53,11 +53,7 @@
             
             UIImageOrientation orientation = [self pin_UIImageOrienationFromImageSource:imageSourceRef];
             
-            
-            
             decodedImage = [self pin_decodedImageWithCGImageRef:imageRef orientation:orientation];
-            
-            
             
             CGImageRelease(imageRef);
         }
@@ -72,7 +68,6 @@
 {
     return [self pin_decodedImageWithCGImageRef:imageRef orientation:UIImageOrientationUp];
 }
-
 
 + (UIImage *)pin_decodedImageWithCGImageRef:(CGImageRef)imageRef orientation:(UIImageOrientation) orientation
 {
--- a/Pod/Classes/Categories/UIImage+DecodedImage.m
+++ b/Pod/Classes/Categories/UIImage+DecodedImage.m
@@ -100,6 +100,7 @@
         decodedImage = [UIImage imageWithCGImage:newImage scale:1.0 orientation:orientation];
         
         CGImageRelease(newImage);
+        CGContextRelease(ctx);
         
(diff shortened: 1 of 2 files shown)


Pod/Classes/Categories/UIImage+DecodedImage.m
```diff
@@ -46,10 +46,18 @@ + (UIImage *)pin_decodedImageWithData:(NSData *)data skipDecodeIfPossible:(BOOL)
...
+    
     if (imageSourceRef) {
         CGImageRef imageRef = CGImageSourceCreateImageAtIndex(imageSourceRef, 0, (CFDictionaryRef)@{(NSString *)kCGImageSourceShouldCache : (NSNumber *)kCFBooleanFalse});
         if (imageRef) {
-            decodedImage = [self pin_decodedImageWithCGImageRef:imageRef];
+            
+            UIImageOrientation orientation = [self pin_UIImageOrienationFromImageSource:imageSourceRef];
+            
+            
+
```
jparise: Mind scaling back on some of the newline / whitespace additions?

Pod/Classes/Categories/UIImage+DecodedImage.m
```diff
@@ -69,22 +83,91 @@ + (UIImage *)pin_decodedImageWithCGImageRef:(CGImageRef)imageRef
...
+        
     } else {
-        decodedImage = [UIImage imageWithCGImage:imageRef];
+        decodedImage = [UIImage imageWithCGImage:imageRef scale:1.0 orientation:orientation];
     }
     
     return decodedImage;
 }
 
++(UIImageOrientation) pin_UIImageOrienationFromImageSource:(CGImageSourceRef)imageSourceRef {
```
jparise: I would consider making this a static helper function in the implementation file.

Pod/Classes/Categories/UIImage+DecodedImage.m
```diff
@@ -69,22 +83,91 @@ + (UIImage *)pin_decodedImageWithCGImageRef:(CGImageRef)imageRef
...
+    
+    //Use UIGraphicsBeginImageContext parameters from docs: https://developer.apple.com/library/ios/documentation/UIKit/Reference/UIKitFunctionReference/#//apple_ref/c/func/UIGraphicsBeginImageContextWithOptions
+    CGContextRef ctx = CGBitmapContextCreate(NULL, imageSize.width, imageSize.height,
+                                             8,
+                                             0,
+                                             colorspace,
+                                             info);
+    
+    CGColorSpaceRelease(colorspace);
+
```
garrettmoon: Is there an advantage to creating a bitmap context instead of letting UIMethod create one?

Pod/Classes/Categories/UIImage+DecodedImage.m
waltflanagan: yea, late on a friday and i forgot to clean up after myself.

Pod/Classes/Categories/UIImage+DecodedImage.m
waltflanagan: Mostly it's to be able to use `UIImage imageWithCGImage:scale:orientation:` so that we can let the UIImage plumbing handling all the transforms related to the orientation.  

Unfortunately the UIGraphics methods only give us access to an opaque CGContext and a resulting UIImage and there is no way to specify the orientation of a UIImage after it's been created.

If you want to continue using the UIGraphics methods and rotate by hand here it can be done, but I don't think there will be any performance benefits and the number of transforms and case statements dealing with that was making my head hurt.

garrettmoon: That all makes great sense.

I asked the question (well, first, because I didn't understand why this was necessary, now I do) because I have a vague memory that creating bitmaps of width not being some multiple make the GPU unhappy (and that using UIGraphicsBeginImageContextWithOptions took care of it for you). But I can't find any reference / documentation around this. I'll dig some more, but this seems like the correct direction.

waltflanagan: I think you're referring to the same hazy thing in my head about 16-byte alignment being optimal for bytesPerRow.  I went with passing 0 to allow Core Graphics to automatically determine the bytes needed, but if there is a performance gain to be had, we could compute the bytes per row accordingly here. 

The docs for `CGBitmapContextCreate` say : `The number of bytes of memory to use per row of the bitmap. If the data parameter is NULL, passing a value of 0 causes the value to be calculated automatically.`

garrettmoon:
This is awesome! Thanks for taking the time to build this. Also of note, iOS 8 looks to have added

kCGImagePropertyOrientationDown
kCGImagePropertyOrientationDownMirrored
kCGImagePropertyOrientationLeft
kCGImagePropertyOrientationLeftMirrored
kCGImagePropertyOrientationRight
kCGImagePropertyOrientationRightMirrored
kCGImagePropertyOrientationUp
kCGImagePropertyOrientationUpMirrored

Could you check if those are defined and if not define them? Then use those instead of one two three, etc.? You can see this in CGImageProperties.h

waltflanagan:
Those constants are defined a CF_ENUM.  Do you have any recommendations on how to determine whether those exist already?  A bit of googling has me stumped.   Ideally we can find something better than checking the SDK version, but from my research it's not looking promising.

As a fallback, i could add an inline comment for each case:
 `case 1: /*kCGImagePropertyOrientationUp*/`

garrettmoon:
Hmm, I think you're right, I'm not sure of a good way either. Lets go with the comments for now. Thanks for doing this!

garrettmoon:
Awesome thanks for your work on this!

Issue #41, opened 2015-08-16
deprecation warning when deployment target is ios8


Sega-Zero:
> /Pods/PINRemoteImage/Pod/Classes/PINRemoteImageManager.m:1245:19: 'threadPriority' is deprecated: first deprecated in iOS 8.0

garrettmoon:
There's a runtime check to make sure it only gets called on iOS 7 or below. Maybe we should wrap in a pragma warning suppressor.

Sega-Zero:
Suppressing warnings is not a good way. I think a better solution would be wrapping in 
```objective-c
#if  __IPHONE_OS_VERSION_MIN_REQUIRED < __IPHONE_8_0
//code
#endif
```
so this code will never gets included in ios8+ apps

garrettmoon:
@Sega-Zero Hmm, but what if you want to support iOS 7 & 8? It only compiles one way or the other, thus the need for a runtime check?

Sega-Zero:
If deployment target is lower than ``_IPHONE_8_0``, the code will be included and the deprecation warning will not be shown. That code will not throw an exception, for now it is just a warning, that ``threadPriority`` should no longer be set manually. There is a better way for this in ios8: the new ``qualityOfService`` property.

garrettmoon:
We were already using qualityOfService with a runtime check :) I've submitted a PR that addresses the warning.

PR #42, opened 2015-08-17
Fix threadPriority deprecation warning

garrettmoon:
Addresses #41


jparise:
:+1:  for `/(compile|run)time/` support.

PR #46, opened 2015-08-26
Fix race conditions in progress generation

garrettmoon:
There was a race condition which existed in which completion would be called before progress was completed. This sets up a serial queue for every download task so progress and completion happen in the correct ordering.

### DIFF
--- a/Pod/Classes/PINRemoteImageManager.m
+++ b/Pod/Classes/PINRemoteImageManager.m
@@ -831,10 +831,7 @@ typedef void (^PINRemoteImageManagerDataCompletion)(NSData *data, NSError *error
                 NSData *data = task.progressImage.data;
             [strongSelf unlock];
             
-            //if task has been removed, no point in calling completion
-            if (task) {
-                completion(data, error);
-            }
+            completion(data, error);
         }
     }];
     
@@ -1014,23 +1011,20 @@ typedef void (^PINRemoteImageManagerDataCompletion)(NSData *data, NSError *error
     
     [progressiveImage updateProgressiveImageWithData:data expectedNumberOfBytes:[dataTask countOfBytesExpectedToReceive]];
 
-    __weak typeof(self) weakSelf = self;
-    [_concurrentOperationQueue pin_addOperationWithQueuePriority:PINRemoteImageManagerPriorityLow block:^{
-        typeof(self) strongSelf = weakSelf;
-        [strongSelf lock];
-            PINRemoteImageDownloadTask *task = [strongSelf.tasks objectForKey:[strongSelf cacheKeyForURL:[[dataTask originalRequest] URL] processorKey:nil]];
-        [strongSelf unlock];
-        
-        if (hasProgressBlocks && [[strongSelf class] isiOS8OrGreater]) {
+    if (hasProgressBlocks && [[self class] isiOS8OrGreater]) {
+        __weak typeof(self) weakSelf = self;
+        [_concurrentOperationQueue pin_addOperationWithQueuePriority:PINRemoteImageManagerPriorityLow block:^{
+            typeof(self) strongSelf = weakSelf;
             UIImage *progressImage = [progressiveImage currentImage];
             if (progressImage) {
                 [strongSelf lock];
+                    PINRemoteImageDownloadTask *task = [strongSelf.tasks objectForKey:[strongSelf cacheKeyForURL:[[dataTask originalRequest] URL] processorKey:nil]];
                     task = [strongSelf.tasks objectForKey:[strongSelf cacheKeyForURL:[[dataTask originalRequest] URL] processorKey:nil]];
                     [task callProgressWithQueue:_callbackQueue withImage:progressImage];
                 [strongSelf unlock];
             }
-        }
-    }];
+        }];
+    }
 }
 
 - (void)didCompleteTask:(NSURLSessionTask *)task withError:(NSError *)error
--- a/Pod/Classes/PINRemoteImageManager.m
+++ b/Pod/Classes/PINRemoteImageManager.m
@@ -1019,8 +1019,7 @@ typedef void (^PINRemoteImageManagerDataCompletion)(NSData *data, NSError *error
             if (progressImage) {
                 [strongSelf lock];
                     PINRemoteImageDownloadTask *task = [strongSelf.tasks objectForKey:[strongSelf cacheKeyForURL:[[dataTask originalRequest] URL] processorKey:nil]];
-                    task = [strongSelf.tasks objectForKey:[strongSelf cacheKeyForURL:[[dataTask originalRequest] URL] processorKey:nil]];
-                    [task callProgressWithQueue:_callbackQueue withImage:progressImage];
+                    [task callProgressWithQueue:strongSelf.callbackQueue withImage:progressImage];
                 [strongSelf unlock];
             }
         }];
--- a/Pod/Classes/PINRemoteImageManager.m
+++ b/Pod/Classes/PINRemoteImageManager.m
@@ -1015,7 +1015,8 @@ typedef void (^PINRemoteImageManagerDataCompletion)(NSData *data, NSError *error
             UIImage *progressImage = [progressiveImage currentImage];
             if (progressImage) {
                 [strongSelf lock];
-                    PINRemoteImageDownloadTask *task = [strongSelf.tasks objectForKey:[strongSelf cacheKeyForURL:[[dataTask originalRequest] URL] processorKey:nil]];
+                    NSString *cacheKey = [strongSelf cacheKeyForURL:[[dataTask originalRequest] URL] processorKey:nil];
+                    PINRemoteImageDownloadTask *task = strongSelf.tasks[cacheKey];
                     [task callProgressWithQueue:strongSelf.callbackQueue withImage:progressImage];
                 [strongSelf unlock];
             }
--- a/Pod/Classes/PINRemoteImageManager.m
+++ b/Pod/Classes/PINRemoteImageManager.m
@@ -93,7 +93,6 @@ typedef void (^PINRemoteImageManagerDataCompletion)(NSData *data, NSError *error
 
 @interface PINRemoteImageManager () <PINURLSessionManagerDelegate>
 {
-    dispatch_queue_t _concurrentQueue;
     dispatch_queue_t _callbackQueue;
     NSLock *_lock;
     NSOperationQueue *_concurrentOperationQueue;
@@ -142,7 +141,6 @@ typedef void (^PINRemoteImageManagerDataCompletion)(NSData *data, NSError *error
         self.cache = [self defaultImageCache];
         NSURLSessionConfiguration *configuration = [NSURLSessionConfiguration defaultSessionConfiguration];
         
-        _concurrentQueue = dispatch_queue_create("PINRemoteImageManagerConcurrentQueue", DISPATCH_QUEUE_CONCURRENT);
         _callbackQueue = dispatch_queue_create("PINRemoteImageManagerCallbackQueue", DISPATCH_QUEUE_CONCURRENT);
         _lock = [[NSLock alloc] init];
         _lock.name = @"PINRemoteImageManager";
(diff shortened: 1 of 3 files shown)


Pod/Classes/PINRemoteImageManager.m
```diff
@@ -831,7 +831,10 @@ - (PINDataTaskOperation *)downloadDataWithURL:(NSURL *)url
                 NSData *data = task.progressImage.data;
             [strongSelf unlock];
             
-            completion(data, error);
+            //if task has been removed, no point in calling completion
```
jparise: We could also call the completion block with a specific "task remove" error. I'm not which approach is more idiomatic, but it seems friendlier to always call a completion block if one is provided so that the caller can take care of any housekeeping.

Pod/Classes/PINRemoteImageManager.m
```diff
@@ -1010,15 +1013,24 @@ - (void)didReceiveData:(NSData *)data forTask:(NSURLSessionDataTask *)dataTask
...
-                task = [self.tasks objectForKey:[self cacheKeyForURL:[[dataTask originalRequest] URL] processorKey:nil]];
-                [task callProgressWithQueue:_callbackQueue withImage:progressImage];
-            [self unlock];
+
+    __weak typeof(self) weakSelf = self;
+    [_concurrentOperationQueue pin_addOperationWithQueuePriority:PINRemoteImageManagerPriorityLow block:^{
+        typeof(self) strongSelf = weakSelf;
+        [strongSelf lock];
+            PINRemoteImageDownloadTask *task = [strongSelf.tasks objectForKey:[strongSelf cacheKeyForURL:[[dataTask originalRequest] URL] processorKey:nil]];
+        [strongSelf unlock];
```
jparise: Is this locked section needed? Two things:

 - This work looks unnecessary if the `hasProgressBlocks && [[strongSelf class] isiOS8OrGreater]` test below fails. 
 - This dictionary lookup looks identical to the one below.

garrettmoon: Yeaaah, I need to clean this up, thanks!

Pod/Classes/PINURLSessionManager.m
```diff
@@ -64,19 +67,37 @@ - (void)unlock
 
 - (void)URLSession:(NSURLSession *)session dataTask:(NSURLSessionDataTask *)dataTask didReceiveData:(NSData *)data
 {
-    [self.delegate didReceiveData:data forTask:dataTask];
+    [self lock];
+    dispatch_queue_t delegateQueue = self.delegateQueues[@(dataTask.taskIdentifier)];
+    [self unlock];
+    
+    __weak typeof(self) weakSelf = self;
+    dispatch_async(delegateQueue, ^{
```
jparise: What happens here if `delegateQueue` is nil?

Pod/Classes/PINURLSessionManager.m
```diff
@@ -42,10 +44,11 @@ - (void)invalidateSessionAndCancelTasks
...
     [self lock];
-    NSURLSessionDataTask *dataTask = [self.session dataTaskWithRequest:request];
-    if (completionHandler) {
-        [self.completions setObject:completionHandler forKey:@(dataTask.taskIdentifier)];
-    }
+        NSURLSessionDataTask *dataTask = [self.session dataTaskWithRequest:request];
+        if (completionHandler) {
+            [self.completions setObject:completionHandler forKey:@(dataTask.taskIdentifier)];
+        }
+        [self.delegateQueues setObject:dispatch_queue_create([[NSString stringWithFormat:@"PINURLSessionManager delegate queue - %ld", (unsigned long)dataTask.taskIdentifier] UTF8String], DISPATCH_QUEUE_SERIAL) forKey:@(dataTask.taskIdentifier)];
```
jparise: This would be a little more readable if you create the dispatch queue as a local variable.

Pod/Classes/PINURLSessionManager.m
```diff
@@ -64,19 +67,37 @@ - (void)unlock
...
-        void (^completionHandler)(NSURLResponse *, NSError *) = self.completions[@(task.taskIdentifier)];
-        [self.completions removeObjectForKey:@(task.taskIdentifier)];
+    dispatch_queue_t delegateQueue = self.delegateQueues[@(task.taskIdentifier)];
     [self unlock];
-    if (completionHandler) {
-        completionHandler(task.response, error);
-    }
+    
+    __weak typeof(self) weakSelf = self;
+    dispatch_async(delegateQueue, ^{
```
jparise: Same here. Any chance that we wouldn't be able to find a queue for this task?

Pod/Classes/PINURLSessionManager.m
garrettmoon: delegateQueue should never be nil.

Pod/Classes/PINURLSessionManager.m
garrettmoon: No, delegate queue is created with the task and only removed in the following dispatch. At least that's *supposed* to be how it works.

Pod/Classes/PINRemoteImageManager.m
```diff
@@ -1010,14 +1010,20 @@ - (void)didReceiveData:(NSData *)data forTask:(NSURLSessionDataTask *)dataTask
...
-        }
+        __weak typeof(self) weakSelf = self;
+        [_concurrentOperationQueue pin_addOperationWithQueuePriority:PINRemoteImageManagerPriorityLow block:^{
+            typeof(self) strongSelf = weakSelf;
+            UIImage *progressImage = [progressiveImage currentImage];
+            if (progressImage) {
+                [strongSelf lock];
+                    PINRemoteImageDownloadTask *task = [strongSelf.tasks objectForKey:[strongSelf cacheKeyForURL:[[dataTask originalRequest] URL] processorKey:nil]];
+                    task = [strongSelf.tasks objectForKey:[strongSelf cacheKeyForURL:[[dataTask originalRequest] URL] processorKey:nil]];
+                    [task callProgressWithQueue:_callbackQueue withImage:progressImage];
```
jparise: `strongSelf.callbackQueue` here so we don't implicitly create an strong ivar reference.

Pod/Classes/PINRemoteImageManager.m
```diff
@@ -1010,14 +1010,20 @@ - (void)didReceiveData:(NSData *)data forTask:(NSURLSessionDataTask *)dataTask
...
-            [self unlock];
-        }
+        __weak typeof(self) weakSelf = self;
+        [_concurrentOperationQueue pin_addOperationWithQueuePriority:PINRemoteImageManagerPriorityLow block:^{
+            typeof(self) strongSelf = weakSelf;
+            UIImage *progressImage = [progressiveImage currentImage];
+            if (progressImage) {
+                [strongSelf lock];
+                    PINRemoteImageDownloadTask *task = [strongSelf.tasks objectForKey:[strongSelf cacheKeyForURL:[[dataTask originalRequest] URL] processorKey:nil]];
+                    task = [strongSelf.tasks objectForKey:[strongSelf cacheKeyForURL:[[dataTask originalRequest] URL] processorKey:nil]];
```
jparise: Is this the same line twice?

garrettmoon: Yes :(

Kurry:
If you create a serial queue for every download task, then you should remove the completion queue property, because currently its not being used. 

```objective-c
@property (atomic, strong) dispatch_queue_t completionQueue;
```

Previously, the method should have looked something like this

```objective-c
- (void)URLSession:(NSURLSession *)session task:(NSURLSessionTask *)task didCompleteWithError:(NSError *)error
{
    [self.delegate didCompleteTask:task withError:error];
    [self lock];
        void (^completionHandler)(NSURLResponse *, NSError *) = self.completions[@(task.taskIdentifier)];
        [self.completions removeObjectForKey:@(task.taskIdentifier)];
    [self unlock];
    
    dispatch_async(self.completionQueue ?: dispatch_get_main_queue(), ^{
        if (completionHandler) {
            completionHandler(task.response, error);
        }
    });
}
```
but with the addition of the delegate queues, then the completion queue should just be removed all together?

garrettmoon:
Hey @Kurry I'm not sure I understand. Do you mean the operation queue? There is no completion queue in PINURLSessionManager. The operation queue is used by the NSURLSession.

Kurry:
@garrettmoon https://github.com/pinterest/PINRemoteImage/blob/master/Pod/Classes/PINURLSessionManager.h#L27

```objective-c
@interface PINURLSessionManager : NSObject

- (instancetype)initWithSessionConfiguration:(NSURLSessionConfiguration *)configuration;

- (NSURLSessionDataTask *)dataTaskWithRequest:(NSURLRequest *)request completionHandler:(void (^)(NSURLResponse *response, NSError *error))completionHandler;

- (void)invalidateSessionAndCancelTasks;

@property (atomic, strong) dispatch_queue_t completionQueue;
@property (atomic, weak) id <PINURLSessionManagerDelegate> delegate;

@end
```

https://github.com/pinterest/PINRemoteImage/blob/master/Pod/Classes/PINRemoteImageManager.m#L160

```objective-c
        _concurrentQueue = dispatch_queue_create("PINRemoteImageManagerConcurrentQueue", DISPATCH_QUEUE_CONCURRENT);

...
        self.sessionManager = [[PINURLSessionManager alloc] initWithSessionConfiguration:configuration];
        self.sessionManager.completionQueue = _concurrentQueue;
        self.sessionManager.delegate = self;
```

Right?

garrettmoon:
Ah! I wasn't looking at the public interface, you're right, good catch! I'll update the PR.

Pod/Classes/PINRemoteImageManager.m
```diff
@@ -1010,14 +1007,19 @@ - (void)didReceiveData:(NSData *)data forTask:(NSURLSessionDataTask *)dataTask
...
-                [task callProgressWithQueue:_callbackQueue withImage:progressImage];
-            [self unlock];
-        }
+        __weak typeof(self) weakSelf = self;
+        [_concurrentOperationQueue pin_addOperationWithQueuePriority:PINRemoteImageManagerPriorityLow block:^{
+            typeof(self) strongSelf = weakSelf;
+            UIImage *progressImage = [progressiveImage currentImage];
+            if (progressImage) {
+                [strongSelf lock];
+                    PINRemoteImageDownloadTask *task = [strongSelf.tasks objectForKey:[strongSelf cacheKeyForURL:[[dataTask originalRequest] URL] processorKey:nil]];
```
jparise: This could be a little more readable if we pull the key generation out into a local variable:

```objc
NSString *cacheKey = [strongSelf cacheKeyForURL:[[dataTask originalRequest] URL] processorKey:nil];
PINRemoteImageDownloadTask *task = strongSelf.tasks[cacheKey];
```

Pod/Classes/PINURLSessionManager.m
```diff
@@ -42,10 +44,12 @@ - (void)invalidateSessionAndCancelTasks
...
-    NSURLSessionDataTask *dataTask = [self.session dataTaskWithRequest:request];
-    if (completionHandler) {
-        [self.completions setObject:completionHandler forKey:@(dataTask.taskIdentifier)];
-    }
+        NSURLSessionDataTask *dataTask = [self.session dataTaskWithRequest:request];
+        if (completionHandler) {
+            [self.completions setObject:completionHandler forKey:@(dataTask.taskIdentifier)];
+        }
+        dispatch_queue_t delegateQueue = dispatch_queue_create([[NSString stringWithFormat:@"PINURLSessionManager delegate queue - %ld", (unsigned long)dataTask.taskIdentifier] UTF8String], DISPATCH_QUEUE_SERIAL);
+        [self.delegateQueues setObject:delegateQueue forKey:@(dataTask.taskIdentifier)];
```
jparise: Similarly, for better readability:

```objc
NSString *queueName = [NSString stringWithFormat:@"PINURLSessionManager delegate queue - %ld", (unsigned long)dataTask.taskIdentifier];
self.delegateQueues[@(dataTask.taskIdentifier)] = dispatch_queue_create([queueName UTF8String], DISPATCH_QUEUE_SERIAL);
```

Pod/Classes/PINURLSessionManager.m
```diff
@@ -64,19 +68,37 @@ - (void)unlock
 
 - (void)URLSession:(NSURLSession *)session dataTask:(NSURLSessionDataTask *)dataTask didReceiveData:(NSData *)data
 {
-    [self.delegate didReceiveData:data forTask:dataTask];
+    [self lock];
+    dispatch_queue_t delegateQueue = self.delegateQueues[@(dataTask.taskIdentifier)];
```
jparise: Indent for the locked scope like you do elsewhere?

jparise:
I added a few code formatting comments, but I think this looks good overall.

jparise:
:+1:

PR #49, opened 2015-09-03
Pass through all the variables :/

garrettmoon:
Image processing wouldn't work if called off the main thread. Turns out calling this off the main thread isn't exactly common, unless you're using ASDK :)


jparise:
:+1:

Issue #23, opened 2015-08-03
Image not found but error is (null)?


donnywals:
When I attempt to load an image at an invalid URL and the completion block is called I would expect the error to contain information about why the image didn't load. Instead both the error and image contain `(null)`. Shouldn't a 404 result in a valid error?

console output when printing result:
```
<PINRemoteImageManagerResult: 0x7ff2996d5e40>image: (null)
animatedImage: (null)
requestDuration: 0.366901
error: (null)
resultType: 3
UUID: <__NSConcreteUUID 0x7ff29ba06db0> 8264EB5B-4ADD-466E-9033-2A11BFDA2524
```

garrettmoon:
Yup, it probably should!

alexvimpekable:
This also the case if url is nil

PR #51, opened 2015-09-10
Bugfix Image Not Found But Error null

Kurry:
This pull request resolves https://github.com/pinterest/PINRemoteImage/issues/23. 

- The reason why a `nil` url did not return an error was because it is handled through an early return method, therefore it does got go through the URL loading system and does not populate the `NSError`.

- The reason why a 404 response does not return in error, is because the [documentation](https://developer.apple.com/library/ios/documentation/Foundation/Reference/NSURLSessionTaskDelegate_protocol/#//apple_ref/occ/intfm/NSURLSessionTaskDelegate/URLSession:task:didCompleteWithError:) states: 

`Server errors are not reported through the error parameter. The only errors your delegate receives through the error parameter are client-side errors, such as being unable to resolve the hostname or connect to the host.`

Which is confusing because if you create a `NSURLSessionTask`, the completion handler of a request which has a `404` response does have the error correctly populated. 

- Empty URLs were handled correctly prior to this pull request, which I determined after it satisfied the tests that I wrote. 

Note: To generate the errors, I just logged what the `NSError` is if you create an analogous `NSURLSessionTask` with a request with the corresponding url, and had the new errors match that.

### DIFF
--- a/Pod/Classes/PINRemoteImageManager.m
+++ b/Pod/Classes/PINRemoteImageManager.m
@@ -627,7 +627,7 @@ typedef void (^PINRemoteImageManagerDataCompletion)(NSData *data, NSError *error
     
     if (completion && ((image || animatedImage) || (url == nil))) {
         //If we're on the main thread, special case to call completion immediately
-        NSError *error;
+        NSError *error = nil;
         if (!url) {
             error = [NSError errorWithDomain:NSURLErrorDomain
                                         code:NSURLErrorUnsupportedURL
--- a/Pod/Classes/PINRemoteImageManager.m
+++ b/Pod/Classes/PINRemoteImageManager.m
@@ -627,11 +627,17 @@ typedef void (^PINRemoteImageManagerDataCompletion)(NSData *data, NSError *error
     
     if (completion && ((image || animatedImage) || (url == nil))) {
         //If we're on the main thread, special case to call completion immediately
+        NSError *error;
+        if (!url) {
+            error = [NSError errorWithDomain:NSURLErrorDomain
+                                        code:NSURLErrorUnsupportedURL
+                                    userInfo:@{ NSLocalizedDescriptionKey : @"unsupported URL" }];
+        }
         if (allowEarlyReturn && [NSThread isMainThread]) {
             completion([PINRemoteImageManagerResult imageResultWithImage:image
                                                           animatedImage:animatedImage
                                                           requestLength:0
-                                                                  error:nil
+                                                                  error:error
                                                              resultType:resultType
                                                                    UUID:nil]);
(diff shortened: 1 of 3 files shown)


Pod/Classes/PINRemoteImageManager.m
```diff
@@ -627,11 +627,17 @@ - (BOOL)earlyReturnWithOptions:(PINRemoteImageManagerDownloadOptions)options url
     
     if (completion && ((image || animatedImage) || (url == nil))) {
         //If we're on the main thread, special case to call completion immediately
+        NSError *error;
+        if (!url) {
+            error = [NSError errorWithDomain:NSURLErrorDomain
+                                        code:NSURLErrorUnsupportedURL
```
jparise: Wouldn't `NSURLErrorBadURL` be more accurate in this case?

Pod/Classes/PINRemoteImageManager.m
```diff
@@ -627,11 +627,17 @@ - (BOOL)earlyReturnWithOptions:(PINRemoteImageManagerDownloadOptions)options url
     
     if (completion && ((image || animatedImage) || (url == nil))) {
         //If we're on the main thread, special case to call completion immediately
+        NSError *error;
```
jparise: Initialize this to `nil`.

Pod/Classes/PINRemoteImageManager.m
Kurry:
```objective-c
    NSURLSessionDownloadTask *task = [[NSURLSession sharedSession] downloadTaskWithRequest:[NSURLRequest requestWithURL:[NSURL URLWithString:nil]] completionHandler:^(NSURL *location, NSURLResponse *response, NSError *error) {
        NSLog(@"%@", error);
    }];
    [task resume];
```

Produces this:
```
2015-09-10 18:31:48.154 PINRemoteImage[13512:130086] Error Domain=NSURLErrorDomain Code=-1002 "unsupported URL" UserInfo=0x7f9d637e5870 {NSLocalizedDescription=unsupported URL, NSUnderlyingError=0x7f9d6505ecf0 "The operation couldn’t be completed. (kCFErrorDomainCFNetwork error -1002.)"}
```

Pod/Classes/PINRemoteImageManager.m
Kurry: thanks

garrettmoon:
Awesome, thank you for doing this!

jparise:
And thanks especially for the tests! :)

Kurry:
Thanks for the merge! No problem, just trying to keep the mind active while being unemployed haha.

Issue #43, opened 2015-08-18
Support for custom HTTP request headers


JaredCrawford:
If an application is required to set additional HTTP headers on image requests, I don't see a way to either supply an ```NSURLRequest``` for each image request (with the appropriate headers already set) or to add the headers globally to all requests.

As one solution, ```PINRemoteImageManager``` uses ```[NSURLSessionConfiguration defaultConfiguration]``` in its initialization. If ```PINRemoteImageManager``` offered a way to customize that configuration, that could be configured to add the additional headers to all image requests made through that (shared) instance of ```PINRemoteImageManager```.

garrettmoon:
Good idea! Would be happy to entertain any pull requests that implement this.

PR #53, opened 2015-09-11
added support for custom http request headers

Kurry:
This resolves https://github.com/pinterest/PINRemoteImage/issues/43. 

I think this approach made the most sense to me, which is allowing a single instance to have a custom configuration, but leaving the `sharedImageManager` unchanged. Let me know your thoughts.

### DIFF
--- a/Example/PINRemoteImage Tests/PINRemoteImage_Tests.m
+++ b/Example/PINRemoteImage Tests/PINRemoteImage_Tests.m
@@ -9,6 +9,7 @@
 #import <UIKit/UIKit.h>
 #import <XCTest/XCTest.h>
 #import <PINRemoteImage/PINRemoteImage.h>
+#import <PINRemoteImage/PINURLSessionManager.h>
 #import <PINRemoteImage/UIImageView+PINRemoteImage.h>
 #import <FLAnimatedImage/FLAnimatedImage.h>
 #import <PINCache/PINCache.h>
@@ -16,6 +17,7 @@
 #if DEBUG
 @interface PINRemoteImageManager ()
 
+@property (nonatomic, strong) PINURLSessionManager *sessionManager;
 @property (nonatomic, readonly) NSUInteger totalDownloads;
 
 - (float)currentBytesPerSecond;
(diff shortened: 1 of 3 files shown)


Example/PINRemoteImage Tests/PINRemoteImage_Tests.m
```diff
@@ -12,10 +12,12 @@
 #import <PINRemoteImage/UIImageView+PINRemoteImage.h>
 #import <FLAnimatedImage/FLAnimatedImage.h>
 #import <PINCache/PINCache.h>
+#import "PINURLSessionManager.h"
```
jparise: `#include <PINRemoteImage/PINURLSessionManager.h>`

Pod/Classes/PINRemoteImageManager.m
```diff
@@ -137,10 +137,18 @@ + (instancetype)sharedImageManager
...
+- (instancetype)initWithSessionConfiguration:(NSURLSessionConfiguration *)configuration
+{
     if (self = [super init]) {
         self.cache = [self defaultImageCache];
-        NSURLSessionConfiguration *configuration = [NSURLSessionConfiguration defaultSessionConfiguration];
-        
+        if (!configuration) {
+            configuration = [NSURLSessionConfiguration defaultSessionConfiguration];
+        } else {
+            configuration = configuration;
```
jparise: This `else` class doesn't seem necessary.

Kurry:
@jparise Rebased, updated code based on feedback, and squashed!

Pod/Classes/PINRemoteImageManager.h
```diff
@@ -116,6 +116,13 @@ typedef void(^PINRemoteImageManagerAuthenticationChallenge)(NSURLSessionTask *ta
 @property (nonatomic, readonly) PINCache *cache;
 
 /**
+ Creates and returns a PINRemoteImageManager created with the specified configuration. If configuration is nil, [NSURLSessionConfiguration defaultConfiguration] is used. You specify a custom configuration if you need to configure timeout values, cookie policies, additional HTTP headers, etc.
```
jparise: I think this should be "Create and return" to match the style of the other comments.

jparise:
This looks good to me. Any other comments, @garrettmoon?

Pod/Classes/PINRemoteImageManager.h
Kurry: great thanks, I made the changed and pushed

jparise:
Thanks, @Kurry!

amruta1234:
@Kurry I dont think your change will work. Cause internally  PIN is calling [PINRemoteManager  sharedManager] function and  allowing a single instance to have a custom configuration, but leaving the sharedImageManager unchanged will not help us set any config params.

garrettmoon:
Related to #72

PR #58, opened 2015-09-14
made changes so FLAnimatedImage is optional / updated PINCache / fix issue with tag not updated

Kurry:
This merge request addresses multiple issues:

1) This fixes a bug where the `PINRemoteImage.podspec` git tag was not updated, when the pod version was updated to `1.2`. 

This error occurred here: https://github.com/pinterest/PINRemoteImage/commit/dcf1fa695b6bc1e26bd36e512a6d47c129d52958

2) Updates PinCache in the Example/Pods folder to reflect upgrade from this commit: https://github.com/pinterest/PINRemoteImage/commit/c00f8ffc9596c32a1ce96446a181cb090ef001b3

3) Makes FLAnimatedImage optional, but is included as part of the default subspec.

### DIFF
--- a/Example/PINRemoteImage Tests/PINRemoteImage_Tests.m
+++ b/Example/PINRemoteImage Tests/PINRemoteImage_Tests.m
@@ -11,7 +11,9 @@
 #import <PINRemoteImage/PINRemoteImage.h>
 #import <PINRemoteImage/PINURLSessionManager.h>
 #import <PINRemoteImage/UIImageView+PINRemoteImage.h>
+#if USE_FLANIMATED_IMAGE
 #import <FLAnimatedImage/FLAnimatedImage.h>
+#endif
 #import <PINCache/PINCache.h>
 
 #if DEBUG
@@ -136,6 +138,7 @@
     [super tearDown];
 }
 
+#if USE_FLANIMATED_IMAGE
 - (void)testGIFDownload
 {
     dispatch_semaphore_t semaphore = dispatch_semaphore_create(0);
@@ -153,6 +156,7 @@
     XCTAssert(outAnimatedImage && [outAnimatedImage isKindOfClass:[FLAnimatedImage class]], @"Failed downloading animatedImage or animatedImage is not an FLAnimatedImage.");
     XCTAssert(outImage == nil, @"Image is not nil.");
 }
+#endif
 
 - (void)testInitWithNilConfiguration
 {
@@ -427,6 +431,7 @@
     [self waitForExpectationsWithTimeout:[self timeoutTimeInterval] handler:NULL];
 }
 
+#if USE_FLANIMATED_IMAGE
 - (void)testFLAnimatedImageView
 {
     XCTestExpectation *imageSetExpectation = [self expectationWithDescription:@"animatedImageView did not have animated image set"];
@@ -441,6 +446,7 @@
 
     [self waitForExpectationsWithTimeout:[self timeoutTimeInterval] handler:NULL];
 }
+#endif
 
 - (void)testEarlyReturn {
     dispatch_semaphore_t semaphore = dispatch_semaphore_create(0);
@@ -460,6 +466,7 @@
     XCTAssert(image != nil, @"image callback did not occur synchronously.");
 }
 
+#if USE_FLANIMATED_IMAGE
 - (void)testload
 {
     srand([[NSDate date] timeIntervalSince1970]);
@@ -492,6 +499,7 @@
     }
     dispatch_group_wait(group, [self timeoutWithInterval:100]);
 }
+#endif
 
 - (void)testInvalidObject
 {
--- a/Example/Pods/Local Podspecs/PINRemoteImage.podspec.json
+++ b/Example/Pods/Local Podspecs/PINRemoteImage.podspec.json
@@ -1,6 +1,6 @@
 {
   "name": "PINRemoteImage",
-  "version": "1.0",
+  "version": "1.2",
   "summary": "A thread safe, performant, feature rich image fetcher",
   "homepage": "https://github.com/pinterest/PINRemoteImage",
   "license": "Apache 2.0",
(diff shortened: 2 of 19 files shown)


Example/PINRemoteImage Tests/PINRemoteImage_Tests.m
```diff
@@ -10,7 +10,9 @@
 #import <XCTest/XCTest.h>
 #import <PINRemoteImage/PINRemoteImage.h>
 #import <PINRemoteImage/UIImageView+PINRemoteImage.h>
+#if __has_include(<FLAnimatedImage/FLAnimatedImage.h>)
```
jparise: Should we use `USE_FLANIMATED_IMAGE` to control these includes, too? It seems like that would be more consistent and easier to control from the top-level configuration.

Pod/Classes/PINRemoteImageManagerResult.h
```diff
@@ -7,8 +7,13 @@
 //
 
 @import UIKit;
-
+#if __has_include(<FLAnimatedImage/FLAnimatedImage.h>)
 #import <FLAnimatedImage/FLAnimatedImage.h>
+#define USE_FLANIMATED_IMAGE    1
+#else
+#define USE_FLANIMATED_IMAGE    0
+#define FLAnimatedImage         NSObject
```
jparise: I'm not sure I like this idea of this alias, especially because this `#define` will leak outside of the PINRemoteImage library to anything that includes its headers.

Example/PINRemoteImage Tests/PINRemoteImage_Tests.m
Kurry: Yeah, I can change that, that makes sense. I guess I had a logical separation where I thought that the sole responsibility of `USE_FLANIMATED_IMAGE` was the macro for defining the object type of `FLAnimatedImage`.

Pod/Classes/PINRemoteImageManagerResult.h
Kurry: I actually copied that from here: https://github.com/pinterest/PINRemoteImage/blob/master/Pod/Classes/PINRemoteImage.h#L20

I had initially thought I was going to add an `#include PINRemoteImage.h` there to get the define, but I thought that I didn't want to pull in the extra dependencies so I added it again.

jparise: Ah, interesting. I think this might also be a good opportunity to clean up that alias, too. We probably want to apply the `USE_FLANIMATED_IMAGE` condition to all FLAnimatedImage references throughout the library.

I'm not very familiar with the FLAnimatedImage component, however, so let's see what @garrettmoon has to say.

Kurry: ok sounds great

Example/Podfile
```diff
@@ -1,9 +1,9 @@
 target 'PINRemoteImage', :exclusive => true do
-  pod "PINRemoteImage", :path => "../"
+  pod "PINRemoteImage/FLAnimatedImage", :path => "../"
```
garrettmoon: shouldn't this just be FLAnimatedImage?
Also, we shouldn't remove pod "PINRemoteImage", :path => "../", right?

Example/Podfile
```diff
@@ -1,9 +1,9 @@
...
   pod "PINCache"
   pod "libwebp"
 end
 
 target 'PINRemoteImage Tests', :exclusive => true do
-  pod "PINRemoteImage", :path => "../"
+  pod "PINRemoteImage/FLAnimatedImage", :path => "../"
```
garrettmoon: Same here?

Example/Podfile
garrettmoon: Hmm, does this have to do with 'PINRemoteImage/Core'?

Example/Pods/Local Podspecs/PINRemoteImage.podspec.json
```diff
@@ -9,26 +9,48 @@
...
+      ],
+      "public_header_files": "Pod/Classes/**/*.h",
+      "frameworks": [
+        "UIKit",
+        "ImageIO",
+        "CoreImage"
+      ],
+      "dependencies": {
+        "PINCache": [
+          ">=2.0"
```
garrettmoon: This dependency now needs to be 2.1

Example/Pods/Local Podspecs/PINRemoteImage.podspec.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "PINRemoteImage",
-  "version": "1.0",
+  "version": "1.1.2",
```
garrettmoon: I just updated the version to 1.2

Example/Pods/Local Podspecs/PINRemoteImage.podspec.json
```diff
@@ -9,26 +9,48 @@
   },
   "source": {
     "git": "https://github.com/pinterest/PINRemoteImage.git",
-    "tag": 1.0
+    "tag": "1.1.2"
```
garrettmoon: Same here

garrettmoon:
Should USE_FLANIMATED_IMAGE be defined as something by default? Maybe if USE_FLANIMATED_IMAGE is undefined and has_include(FLAnimatedImage etc) is true, then USE_FLANIMATED_IMAGE should be true?

garrettmoon:
@Kurry would love to get this merged in, any chance you could take a look at my comments? I think addressing the version number stuff should allow this to merge without conflicts.

PINRemoteImage.podspec
```diff
@@ -19,12 +19,22 @@ Pod::Spec.new do |s|
 
   s.platform     = :ios, '6.0'
   s.requires_arc = true
+  
+  # Exclude optional FLAnimatedImage module
+  s.default_subspec = 'Core'
```
nap-sam-dean: Will this make the default behaviour of the pod be to exclude FLAnimatedImage instead of including it - will this have the effect of `pod update` breaking clients?

Kurry:
Of course! No problem, I'm moving apartments and stuff and can look at it tomorrow or Friday by the latest. Thanks for the updated comments. Will try to get to it ASAP.

PINRemoteImage.podspec
Kurry: @nap-sam-dean 
1. The default behavior will be to exclude FLAnimatedImage.
2. If you aren't using FLAnimatedImage in your project, running `pod update` should not affect your build.

garrettmoon: @Kurry I think it better that the default option be to include the pod? Everyone who is using it now is assuming that FLAnimatedImage is available.

Is it difficult to turn off if the default option is to include it? I'm not familiar enough with subspecs…

Kurry: @garrettmoon Yeah, I read up on it for hours lol so I'm quite familiar with the subspecs. I can make the default to include it. That makes a lot of sense.

Kurry:
@garrettmoon

In response to your comment:

```
Should USE_FLANIMATED_IMAGE be defined as something by default? Maybe if USE_FLANIMATED_IMAGE is undefined and has_include(FLAnimatedImage etc) is true, then USE_FLANIMATED_IMAGE should be true?
```

```objective-c
#if __has_include(<FLAnimatedImage/FLAnimatedImage.h>)
#define USE_FLANIMATED_IMAGE    1
#else
#define USE_FLANIMATED_IMAGE    0
#define FLAnimatedImage         NSObject
#endif
```

I think the code you wrote here ensures that USE_FLANIMATED_IMAGE is always defined to something by default. You just have to note that we use the macro for both deciding if we should import a framework and when we should map `FLAnimatedImage` to `NSObject`.

Example/Pods/Local Podspecs/PINRemoteImage.podspec.json
Kurry: I figured out why this was happening, after I rebased I noticed this was still happening, and it was because the tag was not updated in the PINRemoteImage.podspec file, so I updated it in my pull request.

Kurry:
@garrettmoon I think should be all set. Let me know if you see anything else.

garrettmoon:
Thanks so much for doing this @Kurry!

PR #62, opened 2015-09-29
Fix crash in PINProgressiveImage

garrettmoon:
[CIFilter outputImage] can return nil, [CIContext createCGImage:fromRect:] requires a non-nil CIImage as input.


rcancro:
Looks good!

PR #65, opened 2015-09-30
Make FLAnimatedImage an optional requirement by using subspecs

### LINKED ISSUE #22: Can we make FLAnimatedImage optional?
For now, `FLAnimatedImage` is set as a dependency in podspec, is it possible to make it optional? Maybe make it a subspec?

nap-sam-dean:
This should fix #22 

To exclude FLAnimatedImage just do `pod 'PINRemoteImage/Core'`. The default behaviour of `pod 'PINRemoteImage'` still includes everything.

There are a few slightly gross imports inside the RemoteImageManager but I can't think of a better way to do this quickly.


nap-sam-dean:
Blast - I've just seen pull request #58 ! And it looks much more succinct that mine!

PR #66, opened 2015-10-01
Add ability to specify default PINRemoteImageManagerDownloadOptions on a per-category basis

kerrmarin:
Hi,

This allows us to specify the default options each category should have. The use case for this is to allow subclasses of UIImageView/UIButton/FLAnimatedImageView to specify the default options they would like applied when setting an image.

I've also fixed the disparity between the podspec's version and the actual tag that was being checked out.

There isn't a CHANGELOG so I didn't add the changes to it, although perhaps it's worth creating one? I'm happy to bundle adding a changelog with this if you agree. It will certainly make it easier to define what the next version released should be.

Look forward to your feedback.

### DIFF
--- a/Pod/Classes/PINRemoteImageCategoryManager.m
+++ b/Pod/Classes/PINRemoteImageCategoryManager.m
@@ -172,7 +172,7 @@
     if (placeholderImage) {
         [view pin_setPlaceholderWithImage:placeholderImage];
     }
-    PINRemoteImageManagerDownloadOptions options = PINRemoteImageManagerDownloadOptionsNone;
+    PINRemoteImageManagerDownloadOptions options = [view pin_defaultOptions];
     if ([view pin_ignoreGIFs]) {
         options |= PINRemoteImageManagerDownloadOptionsIgnoreGIFs;
     }
--- a/Pod/Classes/PINRemoteImageCategoryManager.m
+++ b/Pod/Classes/PINRemoteImageCategoryManager.m
@@ -172,7 +172,14 @@
     if (placeholderImage) {
         [view pin_setPlaceholderWithImage:placeholderImage];
     }
-    PINRemoteImageManagerDownloadOptions options = [view pin_defaultOptions];
+    
+    PINRemoteImageManagerDownloadOptions options;
+    if([view respondsToSelector:@selector(pin_defaultOptions)]) {
+        options = [view pin_defaultOptions];
+    } else {
+        options = PINRemoteImageManagerDownloadOptionsNone;
+    }
+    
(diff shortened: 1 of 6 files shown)


Pod/Classes/PINRemoteImageCategoryManager.m
```diff
@@ -172,7 +172,7 @@ + (void)setImageOnView:(id <PINRemoteImageCategory>)view
     if (placeholderImage) {
         [view pin_setPlaceholderWithImage:placeholderImage];
     }
-    PINRemoteImageManagerDownloadOptions options = PINRemoteImageManagerDownloadOptionsNone;
+    PINRemoteImageManagerDownloadOptions options = [view pin_defaultOptions];
```
garrettmoon: Since this will break any existing subclasses that don't implement this, can you make it optional in the protocol and check that the view responds to selector?

kerrmarin: Hey, thanks for getting back to me. I'm not sure I follow -- what subclasses do you mean? This won't break any subclasses of UIImageView/UIButton/FLAnimatedImageView. This is a perfectly valid subclass:

```
@import UIKit;
#import "UIImageView+PINRemoteImage.h"

@interface UIImageViewSubclass : UIImageView
@end
@implementation UIImageViewSubclass

//Not implementing pin_defaultOptions, but I want to override GIF ignoring
- (BOOL)pin_ignoreGIFs {
    return YES;
}

@end
```

I guess it would break any classes outside of this library that conform to the `PINRemoteImageCategory` but I'm not sure if its a good idea to support that?

garrettmoon: Yup, my point is it would break any subclasses which conform with the protocol. The reason it's a protocol is to specifically allow categories outside the library :) For example, we have a category to couple AsyncDisplayKit's ASImageNode to PINRemoteImage.

kerrmarin: I'm finding it quite hard to come up with a good reason! I guess its personal preference. For example, I wouldn't expect people to create categories that conformed to `PINRemoteImageCategory`. I find that if you have the need for that, then make a PR. If the extension brings in other dependencies, then perhaps create a subpod for it along with your feature <cough> ASImageNode <cough> ;). The current way, it just means that if there's a situation where you _expect_ every category to implement a method, its either a major version bump or another `respondsToSelector` wrap-around.

I've made the changes to make it `@optional`. Thanks! :)

garrettmoon:
Cool, thanks!

PR #67, opened 2015-10-05
Add launch screen to Example project

kerrmarin:
Hey,

No changes to the library itself, just to the Example project. I've added a launch image so that the example project displays correctly in the iPhone 6 [Plus].

**There's also a change to the `.travis.yaml` file to start building using xcode7. If this is not something you want then feel free to reject this PR.**

Kerr


kerrmarin:
This is going to take a little more work than I anticipated -- I'll reopen this when its ready

PR #71, opened 2015-10-11
Return image result type of 'None' if no image is downloaded

### LINKED ISSUE #68: Receiving resultType of PINRemoteImageResultTypeDownload when offline and no image is downloaded
I expect that when I get a result in my completion block with a non-nil `error`, and a nil `image`, the `resultType` should be `PINRemoteImageResultTypeNone` - no image was downloaded, after all.

However, with my iPad in Airplane mode, I get `PINRemoteImageResultTypeDownload`. 

This happens because the result type is [assumed to be a successful download if a cache miss occurs](https://github.com/pinterest/PINRemoteImage/blob/master/Pod/Classes/PINRemoteImageTask.m#L61), regardless of the error state or images existence.

Is this expected behaviour?

adamsp:
Previously the `resultType` being returned was `PINRemoteImageResultTypeDownload` even when the download failed. Now, we check to see if the `image` has been set and if it not then we return `PINRemoteImageResultTypeNone`.

This fixes #68.

### DIFF
--- a/Pod/Classes/PINRemoteImageTask.m
+++ b/Pod/Classes/PINRemoteImageTask.m
@@ -54,11 +54,17 @@
             PINLog(@"calling completion for UUID: %@ key: %@", UUID, strongSelf.key);
             dispatch_async(queue, ^
             {
+                PINRemoteImageResultType result;
+                if (image || animatedImage) {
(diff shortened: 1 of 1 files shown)


garrettmoon:
Awesome, thanks!

PR #74, opened 2015-10-14
So, uh,

garrettmoon:
+ (UIImage * _Nonnull)imageWithCGImage:(CGImageRef _Nonnull)cgImage

says it returns a non-nil value but documentation says:

Return Value
A new image object for the specified Quartz image, or nil if the method could not initialize the image from the specified image reference.

So we're guarding against a nil inputImage


appleguy:
:) looks good, r=appleguy :)

PR #75, opened 2015-10-14
Optionally pass full result to category implementor

bdorfman:
An image-rendering object might want more details of the download than just the image when a download is completed and it needs to render the new result, so optionally give it the whole result object.

Example usage: Our image view subclass renders the setting of the new image as an animated fade-in if the image was newly downloaded and took more than a tenth of a second to complete, so that it transitions nicely from the placeholder image.

### DIFF
--- a/Pod/Classes/PINRemoteImageCategoryManager.m
+++ b/Pod/Classes/PINRemoteImageCategoryManager.m
@@ -195,7 +195,13 @@
                     return;
                 }
                 if (result.image) {
-                    [view pin_updateUIWithImage:result.image animatedImage:nil];
+                    if ([view respondsToSelector:@selector(pin_updateUIWithRemoteImageManagerResult:)]) {
+                        [view pin_updateUIWithRemoteImageManagerResult:result];
+                    }
+                    else {
+                        [view pin_updateUIWithImage:result.image animatedImage:nil];                        
+                    }
+
                 }
             };
             if ([NSThread isMainThread]) {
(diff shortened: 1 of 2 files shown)


Pod/Classes/PINRemoteImageCategoryManager.m
```diff
@@ -195,7 +195,13 @@ + (void)setImageOnView:(id <PINRemoteImageCategory>)view
                     return;
                 }
                 if (result.image) {
-                    [view pin_updateUIWithImage:result.image animatedImage:nil];
+                    if ([view respondsToSelector:@selector(pin_updateUIWithRemoteImageManagerResult:)]) {
```
garrettmoon: Minor thought. Could you cache the result of responds to selector in a BOOL? Maybe BOOL updateWithFullResult = [view respondsToSelector:@selector(pin_updateUIWithRemoteImageManagerResult:)];

garrettmoon:
Nice! Just had one comment.

Pod/Classes/PINRemoteImageCategoryManager.m
bdorfman: Ah, you mean so it doesn't make the same check twice in the same method? Yeah good idea I will make the change.

garrettmoon: Yup, thanks!

bdorfman:
Updated.

Issue #25, opened 2015-08-04
Carthage support


angelolloqui:
It would be nice to have carthage support. Not sure about the exact requirements, but I believe it needs to have a framework target.

garrettmoon:
I'd love to add Carthage support, if you're interested in writing a pull request, I'd happily accept :)

dcaunt:
See [Supporting Carthage for your framework](https://github.com/Carthage/Carthage#supporting-carthage-for-your-framework)

This involves creating an Xcode project with framework targets and sharing the schemes. Having the project file will generally benefit those not using CocoaPods. I may have a stab at this if I get time.

pmcatominey:
Looking into this, it appears we'll need FLAnimatedImage to be available through Carthage first.

I've started playing with this in my fork.

dcaunt:
I have `PINRemoteImage.framework` building locally with Carthage but I don't think it can work for Carthage users until FLAnimatedImage merges in https://github.com/Flipboard/FLAnimatedImage/pull/79.

In the meantime I can PR the framework support with FLAnimatedImage as a submodule which would solve https://github.com/pinterest/PINRemoteImage/issues/30.

garrettmoon:
@dcaunt Awesome! Do you mean #22 ? It'd be great for the project if FLAnimatedImage could be optional!

dcaunt:
Sorry, I'll try to explain this more clearly.

There are three issues here: 
#22 FLAnimatedImage is a non-optional dependency
#25 Carthage support (this issue)
#30 Framework target

Carthage support depends on #30 – we can't support building via Carthage without a framework. In terms of building a PINRemoteImage.framework, we need to add an xcodeproj, and link to FLAnimatedImage somehow. As FLAnimatedImage doesn't support Carthage yet, the best way is to include it as a git submodule and then the framework can be built in Xcode.

In terms of making FLAnimatedImage an optional dependency, I don't know how to achieve this off the top of my head for both CocoaPods and frameworks.

pmcatominey:
@dcaunt Split FLAnimatedImage support into a separate repo, setup to depend on PINRemoteImage for both CocoaPods and Carthage?

dcaunt:
That makes sense, I was looking for a way to keep everything together in one repo but I think a separate repo would be more straightforward.

garrettmoon:
[...]

PR #77, opened 2015-10-15
Adds support for Carthage

bdorfman:
This PR adds Carthage support to PINRemoteImage. It also as a side effect should make it slightly easier for people who want to manually clone/submodule the repo and build the framework to do so. However, I am concerned I may have accidentally broke the existing Cocoapods support since I don't quite understand the Cocoapods flow, so there may be further changes needed in that regard before this can be merged.

This adds a new Xcode project that builds PINRemoteImage.framework. Xcode wants frameworks' umbrella/master header file (the one named FrameworkName.h) to import all public headers in the framework or it throws a warning. To accomplish this I had to move macros out of PINRemoteImage.h into a separate file to resolve circular dependency issues. I also properly guarded the compilation of FLAnimatedImage code around whether or not the library is available, and removed the OS_IPHONE check as this framework does not support OS X anyway (and in order to do so would need a lot more changes than just swapping UIImage for NSImage).

PINCache is added as a Carthage dependency so if you use Carthage that will be taken care of for you automatically. It should theoretically also work with FLAnimatedImage via Carthage if you have that in your parent app's Cartfile, although their Carthage support seems a little broken at the moment so I couldn't test it properly.


garrettmoon:
Sorry I haven't looked at this yet, I'll try and take a look next week.

bdorfman:
No worries, I can just use my own fork for our app so there's no rush.

PINRemoteImage/PINRemoteImage.xcodeproj/project.xcworkspace/xcshareddata/PINRemoteImage.xcscmblueprint
```diff
@@ -0,0 +1,30 @@
...
+  "DVTSourceControlWorkspaceBlueprintVersion" : 204,
+  "DVTSourceControlWorkspaceBlueprintRelativePathToProjectKey" : "PINRemoteImage\/PINRemoteImage.xcodeproj",
+  "DVTSourceControlWorkspaceBlueprintRemoteRepositoriesKey" : [
+    {
+      "DVTSourceControlWorkspaceBlueprintRemoteRepositoryURLKey" : "https:\/\/github.com\/pinterest\/PINCache.git",
+      "DVTSourceControlWorkspaceBlueprintRemoteRepositorySystemKey" : "com.apple.dt.Xcode.sourcecontrol.Git",
+      "DVTSourceControlWorkspaceBlueprintRemoteRepositoryIdentifierKey" : "44B36422A7A7C3AD3DE414BC27AA89E3C0009BBA"
+    },
+    {
+      "DVTSourceControlWorkspaceBlueprintRemoteRepositoryURLKey" : "github.com:foursquare\/PINRemoteImage.git",
```
garrettmoon: Should this file be checked in? Should anything in xcshareddata?

bdorfman: Ah yeah, that is probably a mistake. You need to check in your shared framework scheme (which is in xcshareddata) so that Carthage can find and build it, but not this blueprint file I don't think. I will fix.

garrettmoon:
Thank you so much for doing this! I tested it locally and it seems CocoaPods is still happy :) Going to do a bit more due diligence and then I think we can get this merged!

bdorfman:
No problem. By the way, I learned more about modulemap minutiae since doing this, so I think I can convince it to not require that all public headers be imported by PINCache.h if you would like me to change that back.

FYI, proper support for FLAnimatedImage via Carthage requires them to merge this PR https://github.com/Flipboard/FLAnimatedImage/pull/95 But theoretically if you list FLAnimatedImage in your Cartfile before you list PINRemoteImage, it will build it with animated gif support.

garrettmoon:
If we can at least get the category imports out that might be a good idea?

bdorfman:
Ok, I think everything should work now. Updated gitignore, removed unnecessary xcode files, update the header/modulemap, and added carthage instructions to readme.

I think the travis-ci build is failing because I don't know how to update the example project properly using cocoapods. I tried just doing `pod install` and that changed a LOT of files that I wasn't sure about so I will just leave it up to you to resolve the proper way.

garrettmoon:
Cool! I'll get this merged.

garrettmoon:
Manually merged, thanks so much for doing this!

garrettmoon:
Hey @bdorfman just tried to figure out this whole Carthage thing and I'm receiving the same error as here:

https://github.com/pinterest/PINRemoteImage/issues/25#issuecomment-153346771

Should there be an xcodeproj specifically for Carthage in the root directory with a shared framework scheme? Did it not get checked in?

PR #81, opened 2015-10-19
Placeholder image is set even if url is nil

### LINKED ISSUE #80: Placeholder Image not set when URL is nil.
When using a PinRemoteImageCategory (UIImageView) if you set a placeholderImage and provide an invalid URL. The invalid url causes the UIImageView to set the image to nil removing the placeholder image. 

I would expect the behaviour to instead set the placeholder Image regardless of if the url is a valid or not. 

I can submit a PR with a fix. If this is something you guys would liked fixed.

farhanpatel:
Allow a PinRemoteImageCategory to set a placeholder image even if the provided url does not exist. Fixes #80

### DIFF
--- a/Pod/Classes/PINRemoteImageCategoryManager.m
+++ b/Pod/Classes/PINRemoteImageCategoryManager.m
@@ -164,13 +164,16 @@
     }
     
     [self cancelImageDownloadOnView:view];
-    if (urls == nil || urls.count == 0) {
-        [view pin_clearImages];
-        return;
-    }
-    
+  
     if (placeholderImage) {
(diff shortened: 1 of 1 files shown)


garrettmoon:
@farhanpatel Thanks for the PR!

I'm not sure I understand the issue. With your pull request, if the URL is nil, it will set the placeholder image but then immediately clear it out. How is this different from clearing any existing image out and returning without setting the placeholder?

farhanpatel:
*It will only clear the image if there is no placeholder. If there is a placeholder than the image is not cleared.

garrettmoon:
Ah! Sorry I misread the PR :)

PR #84, opened 2015-10-23
Fix a race condition in PINURLSessionManager

garrettmoon:
The callback queue cannot be concurrent because there's no way to guarantee the
operations happen FIFO.

Also fixed an issue where authentication wouldn't happen if the PINURLSessionManager
delegate didn't respond to didReceiveChallenge:

### DIFF
--- a/Pod/Classes/PINURLSessionManager.m
+++ b/Pod/Classes/PINURLSessionManager.m
@@ -26,6 +26,7 @@
         self.sessionManagerLock = [[NSLock alloc] init];
         self.sessionManagerLock.name = @"PINURLSessionManager";
         self.operationQueue = [[NSOperationQueue alloc] init];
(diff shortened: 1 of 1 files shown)


Pod/Classes/PINURLSessionManager.m
```diff
@@ -26,7 +26,9 @@ - (instancetype)initWithSessionConfiguration:(NSURLSessionConfiguration *)config
         self.sessionManagerLock = [[NSLock alloc] init];
         self.sessionManagerLock.name = @"PINURLSessionManager";
         self.operationQueue = [[NSOperationQueue alloc] init];
```
jparise: Also consider setting the queue's name.

jparise:
:+1:

PR #86, opened 2015-10-27
Guard against nil url in category methods

bdorfman:
The other methods on PINRemoteImageCategoryManager all do this already, but there is one method on both UIImageView and UIButton that doesn't check for nil url before wrapping it in an array.


garrettmoon:
Thanks!

PR #95, opened 2015-11-09
Fix for local image URLs throwing an error.

connor:
Loading a local image URL would throw an error because it is just a `NSURLResponse` instead of an `NSHTTPURLResponse` and thus calling `statusCode` would throw an unrecognized selector error.

A regression from b7f1856d36.

### DIFF
--- a/Pod/Classes/PINURLSessionManager.m
+++ b/Pod/Classes/PINURLSessionManager.m
@@ -102,7 +102,7 @@
     [self lock];
         dispatch_queue_t delegateQueue = self.delegateQueues[@(task.taskIdentifier)];
     [self unlock];
-    if (!error && [task isKindOfClass:[NSHTTPURLResponse class]] && [(NSHTTPURLResponse *)task.response statusCode] == 404) {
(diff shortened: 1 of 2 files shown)


Pod/Classes/PINURLSessionManager.m
```diff
@@ -102,7 +102,8 @@ - (void)URLSession:(NSURLSession *)session task:(NSURLSessionTask *)task didComp
     [self lock];
         dispatch_queue_t delegateQueue = self.delegateQueues[@(task.taskIdentifier)];
     [self unlock];
-    if ([(NSHTTPURLResponse *)task.response statusCode] == 404 && !error) {
+    NSHTTPURLResponse *response = (NSHTTPURLResponse *)task.response;
+    if ([response respondsToSelector:@selector(statusCode)] && [response statusCode] == 404 && !error) {
```
jparise: This is a little odd because you've already cast the response to `NSHTTPURLResponse`, effectively asserting that it should have `statusCode`.

My suggestion:

```objc
if (!error && [task.response respondsToSelector:@selector(statusCode)] && [task.response statusCode] == 404) {
```

PR #99, opened 2015-11-24
CIImage doesn't retain its CGImage :(

garrettmoon:
CIImage wasn't retaining the CGImage returned from UIImage. This patch manually retains it.


jparise:
Ouch. :+1:

PR #105, opened 2015-12-07
Support for sharedImageManager with custom configuration

### LINKED ISSUE #72: Using PINRemoteImageManager's initWithSessionConfiguration
I am unclear on how you are supposed to use the `initWithSessionConfiguration` init on `PINRemoteImageManager`, as there is no way to get your init'd manager into the result of the `sharedImageManager` method, which is used by several methods in `PINRemoteImageCategoryManager`. Possibly I am just not understanding its intended usage, but it seems like you need to either have a way to change/set the return value of sharedImageManager, or refactor `PINRemoteImageCategoryManager` to not rely on a singleton.

bsmith11:
This should resolve #72

### DIFF
--- a/Pod/Classes/PINRemoteImageManager.h
+++ b/Pod/Classes/PINRemoteImageManager.h
@@ -129,6 +129,13 @@ typedef void(^PINRemoteImageManagerAuthenticationChallenge)(NSURLSessionTask *ta
 + (instancetype)sharedImageManager;
 
 /**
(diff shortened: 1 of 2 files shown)


bsmith11:
Looks like the QoS test timed out, but I don't think my changes could affect that one. Any thoughts?

garrettmoon:
I like it!

PR #106, opened 2015-12-08
Update WebP connection to support CocoaPods+Frameworks

Adlai-Holler:
Love this framework! This makes it possible to use WebP with PINRemoteImage with the `use_frameworks!` option. We have to use a preprocessor definition because at compile time we won't have `libwebp` headers accessible.

Let me know if there's anything that needs tweaking. I tested that this worked in the `use_frameworks!` case.


garrettmoon:
Thanks for doing this! And sorry for the long delay in looking at it. Can you take a look at why the webp tests are failing? I assume it has to do with the test projects spec not having webp?

mRs-:
`use_frameworks!` doesn't work with iOS 7

Adlai-Holler:
Sorry for the delay here, I'm totally swamped and I let this slip through the cracks. I'm cutting a release right now, but I'll investigate the failing tests this week.

garrettmoon:
Hey @Adlai-Holler any chance you can look at this soon?

Adlai-Holler:
The party's moved to #129

Issue #92, opened 2015-11-07
 Crashed: PINURLSessionManager Operation Queue :: NSOperation (QOS: LEGACY)


farhanpatel:
We have a similar issue to #91 I thought I'd create a new ticket as it isnt on the exact same line. I'm running the latest code with commit 65d9c582fbaed691940cf4c550054535e4b84246 

The issue happens on ios 8/9 only. I can link you to the crashlytics issue if more info is required. 
```
Thread : Crashed: PINURLSessionManager Operation Queue :: NSOperation 0x149041870 (QOS: LEGACY)
0  libdispatch.dylib              6850538312 dispatch_async + 76
1  libdispatch.dylib              6850538304 dispatch_async + 68
2  PaigeeWorld                    4297916584 -[PINURLSessionManager URLSession:task:didReceiveChallenge:completionHandler:] (PINURLSessionManager.m:79)
3  Foundation                     6509364084 __NSBLOCKOPERATION_IS_CALLING_OUT_TO_A_BLOCK__ + 16
4  Foundation                     6508655008 -[NSBlockOperation main] + 96
5  Foundation                     6508590056 -[__NSOperationInternal _start:] + 604
6  Foundation                     6509373288 __NSOQSchedule_f + 224
7  libdispatch.dylib              6850533032 _dispatch_client_callout + 16
8  libdispatch.dylib              6850582252 _dispatch_queue_drain + 864
9  libdispatch.dylib              6850548140 _dispatch_queue_invoke + 464
10 libdispatch.dylib              6850590140 _dispatch_root_queue_drain + 728
11 libdispatch.dylib              6850589404 _dispatch_worker_thread3 + 112
12 libsystem_pthread.dylib        6852760688 _pthread_wqthread + 1092
```

I've had no luck trying to reproduce it. If you have any thoughts on how I can give it a shot.

garrettmoon:
Looking at it in crashlytics would definitely be more informative.

farhanpatel:
Here is the report. You should be able to access it. http://crashes.to/s/859ce80843f

farhanpatel:
Here is some more info via HockeyApp. It has some info on what the main thread was doing. It crashes on the same spot but not sure if its the exact same issue.
https://gist.github.com/farhanpatel/815df738603cc932cb27

bdorfman:
We are also seeing a similar crash in the didReceiveChallenge callback. I can't quite figure out what is happening. 

[...]

PR #107, opened 2015-12-09
Fix crash in urlSession authentication challenge.

bdorfman:
Implement delegate method for when there is a session-level authentication challenge so that the call isn't forwarded to the task version.
Make sure we always strongSelf inside blocks.

Theoretically fixes https://github.com/pinterest/PINRemoteImage/issues/92


garrettmoon:
Awesome, thank you!

Issue #70, opened 2015-10-09
Ability to track download progress


AdamBCo:
Is there a way to track the download progress of a remote image?

For example, I would like to show a progress bar atop a UIImageView that shows the image is downloading from the web.

Issue #59 referred to it, but I think it was answered in reference to progressive renders as opposed to download percentages.

exherb:
here https://gist.github.com/exherb/e0ad899fe5d3fc7ea2b3

garrettmoon:
@exherb Nice! An easy way to solve the issue without modifying the library any :)

If we were to add this functionality to PINRemoteImage, my suggestion would be to add a progress percentage to PINRemoteImageManagerResult and modify how the progressive progress method calls back to have an option to disable progressive renders and to call back every time it gets data, not just when it has a progressive render.

connor:
Closing this out since #134 added support for it and it's been published in 2.0 release.

VictorZhang2014:
@exherb  Your gist link is not found, have you moved the code snippet to another place?

PR #110, opened 2016-01-02
Add optional block to track download progress of image.

christianselig:
Adds an additional optional block to `PINRemoteImageManager` to allow users to track progress of image downloads in bytes. Referenced in #70 (@exherb's subclass served as a great reference, cheers!).

It's a complicated library, so hopefully I implemented it correctly. :) I was slightly confused about `callbackCompletions` on the image download tasks, where there can be multiple per task (and I'm still not sure I completely understand it) but I simply called the download progress block on any callbackCompletion that had the block.

In the aforementioned issue, @garrettmoon mentioned:

> If we were to add this functionality to PINRemoteImage, my suggestion would be to add a progress percentage to PINRemoteImageManagerResult and modify how the progressive progress method calls back to have an option to disable progressive renders and to call back every time it gets data, not just when it has a progressive render.

Which I totally understand, but personally I chose to do it as an additional block all together, allowing you to not have to necessarily choose between progressive renders and finding out the progress of the download. Furthermore I didn't implement this for loading an image with multiple different quality URLs (don't quite have the time right now and how that would work wasn't immediately intuitive).

As [Alamofire does](https://github.com/Alamofire/Alamofire#uploading-with-progress), due to potential performance issues the block is purposefully not automatically dispatched to the main thread on every invocation, so I added a note in the documentation that the user should ensure that UI updates in the block are dispatched to the main thread.

Phenomenal library by the way, just really needed this feature. :)


Pod/Classes/PINRemoteImageManager.m
```diff
@@ -1058,6 +1114,16 @@ - (void)didReceiveData:(NSData *)data forTask:(NSURLSessionDataTask *)dataTask
                 task.progressImage.progressThresholds = self.progressThresholds;
             }
         }
+    
+        for (PINRemoteImageCallbacks *completion in task.callbackBlocks.allValues) {
```
garrettmoon: I think it might be better to copy the completions and call them outside of the lock?

garrettmoon:
Thanks so much for doing this, it's a great improvement. I'm wondering if we should rename the progress parameter to progressImage?

christianselig:
Sure! I'll give it a go in a bit.

christianselig:
Okay, I think it's complete, @garrettmoon. Names are changed and blocks are called outside of the lock now (also pulled your master branch down and merged it with my feature branch to fix a merge conflict).

garrettmoon:
@christianselig It looks like several files were deleted and re-added in a different place. I'd prefer if we could not check these in as it will rewrite the history on those files.

garrettmoon:
@maicki got this ready to go, @christianselig please let me know if what landed doesn't work for you! Thanks for your help on this!

christianselig:
@garrettmoon Looks awesome, happy to be able to help!

PR #111, opened 2016-01-04
Pass through setUpdateWithProgress: value to prefixed property

maicki:
Fixes the setUpdateWithProgress: method that currently calls itself instead of passing the value through to the prefixed pin_updateWithProgress property. This results into a infinite loop and a crash.


jparise:
Thanks!

Issue #50, opened 2015-09-10
How about support OS X?


crossle:

garrettmoon:
Sounds like a great idea! Shouldn't be too much work, though I think it would be best to make FLAnimatedImage optional first. Happy to accept any pull requests adding support for OS X.

mRs-:
It's done, can be closed

PR #112, opened 2016-01-04
Add OS X support

maicki:
**Please note:** Before you start looking into the iOS and OSX examples please run "pod install" for both of the examples. If I would do a "pod install" and push this the pull request would be pretty big and code review would be pretty hard.

This pull request is a first implementation to add cross platform support to PINRemoteImage (Issue: #50 ). As there are different ways to add cross platform support there are a couple of questions open I want to address below.

**I looked into different options to add cross platform support:**
1st option:
The first option is the way cross platform support is currently implemented, by mapping generic classes like PINImage, PINImageView and PINButton to the equivalent UIKit or AppKit classes. Throughout the framework use these mappings instead of the concrete UIKit / AppKit names. Furthermore we provide a category on NSImage that is adding methods that are not available on NSImage but on UIImage.

2nd option:
The second option to add cross platform support would be to always check if we build for iOS ( __IPHONE_OS_VERSION_MIN_REQUIRED) or OSX (__MAC_OS_X_VERSION_MIN_REQUIRED) and based on that use UIImage or NSImage in code. The downside of this way is that there would be a lot of "if else" throughout the code base that either compiles for iOS or OS X.

3rd option:
The third option would be to instead of using UIImage and NSImage, we use id everywhere in code. In my opinion this would decrease the typesafety tremendously and especially if used with Swift this would be a step back and so I decided against it.

**API changes:**
As I implemented the first way described above, I also did some API changes. The biggest change is that the UIKit (UIImage, UIImageView and UIButton) categories are changed to generic ones that support categories on UIKit and AppKit classes. As result of that I also changed the names of the category files to generic one. This will likely break imports for PINRemoteImage user. I'm happy for every improvements on this case. One thing we could do is to let the UIKit categories and as they are and deprecated them as well as exclude them in the podspec for OSX and in the documentation as well as sample apps only refer to the new generic once.

Another way it could be implemented is to add separate categorie files for every platform. In the podspec we would exclude the UIKit categories in the OSX subspec and exclude AppKit categories in the iOS subspec. The downside from my opinion is that as you can see in the current implementation we would just copy and paste most of the code except using NSImage instead of UIImage and two or three different method calls.

I'm curious about the general opinions and how we should proceed based on the discussion above.

**TODO:**
- [x] Map UIKit / AppKit classes (PINImage, PINImageView, PINButton)
- [x] Rename UIKit categories
- [x] Handle: isiOS8OrGreater
- [x] Add OSX support to podspec file
- [x] Update iOS sample project
- [x] Proper Mac sample project
- [ ] Update readme for usage on OSX


garrettmoon:
@maicki This is awesome! Thank you for putting in the work!

I agree with your choices for mapping to UIKit / AppKit. This is a large enough change that it warrants a 2.0 release I think.

Let me know when you're ready for a review and I'll look in more detail!

maicki:
@garrettmoon Added a better example for OS X. The goal was to have the same features as on iOS and share at least the model (Kitten) classes between the examples.
We could go even a bit more cross platform and share the same view controller between iOS and OS X in the examples. Curious what you think?

Furthermore except the readme update, I think code wise it would be good to have the first look over it. 

Thanks!

garrettmoon:
I'll take a look at this soon. I'd suggest we get the conflicts resolved, but we'll probably want to plan a 2.0 release for this so it may be a bit before we can get it merged in.

maicki:
@garrettmoon The conflict should be resolved, though travis failed. If you look at the details you will see one of the builds succeed and one failed due to a network timeout.

Although it's planned for a 2.0 release I think would be a good time to look over it again as pushed all the latest changed and merged in master + resolved all merge conflict.

garrettmoon:
@maicki the travis build fails sometimes due to timeout, I kicked it off again. I'll try and get this merged today.

garrettmoon:
Thank you so much for taking the time to do this @maicki !

PR #116, opened 2016-01-19
After continually seeing crashes on createCGImage:fromRect: I'm ready

garrettmoon:
to try some speculative fixes. This moves the rendering to a serial queue
A bit disappointing, this will be reverted if it doesn't address the crashes.

If it does… well, we'll have more to work on.


appleguy:
r=me.  I agree with the plan of attack here.  I have discovered some thread safety issues in Apple's frameworks recently, so unfortunately we have to be pretty defensive these days...

PR #120, opened 2016-01-22
Switch to Accelerate for blurring

garrettmoon:
Core Image crashes were still showing up in our stack traces
even after many speculative fixes. Lets see if accelerate does
better.

### DIFF
--- a/Pod/Classes/PINProgressiveImage.m
+++ b/Pod/Classes/PINProgressiveImage.m
@@ -68,9 +68,6 @@
             CFRelease(_imageSource);
         }
     [self.lock unlock];
-#if !PIN_APP_EXTENSIONS
-    [[NSNotificationCenter defaultCenter] removeObserver:self];
-#endif
 }
 
 #pragma mark - public
--- a/Pod/Classes/PINProgressiveImage.m
+++ b/Pod/Classes/PINProgressiveImage.m
@@ -153,7 +153,7 @@
     [self.lock unlock];
 }
 
-- (UIImage *)currentImageBlurred:(BOOL)blurred
+- (UIImage *)currentImageBlurred:(BOOL)blurred maxProgressiveRenderSize:(CGSize)maxProgressiveRenderSize
 {
     [self.lock lock];
         if (self.imageSource == nil) {
@@ -203,6 +203,11 @@
             NSDictionary *jpegProperties = imageProperties[(NSString *)kCGImagePropertyJFIFDictionary];
             NSNumber *isProgressive = jpegProperties[(NSString *)kCGImagePropertyJFIFIsProgressive];
             self.isProgressiveJPEG = jpegProperties && [isProgressive boolValue];
+        }
+    
+        if (self.size.width > maxProgressiveRenderSize.width || self.size.height > maxProgressiveRenderSize.height) {
+            [self.lock unlock];
+            return nil;
         }
         
         float progress = 0;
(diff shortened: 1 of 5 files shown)


Pod/Classes/PINProgressiveImage.m
```diff
@@ -350,59 +301,135 @@ - (CFTimeInterval)estimatedRemainingTime
...
-    UIImage *outputUIImage = nil;
-    
-    CIContext *processingContext = GPUProcessingContext;
-    if (self.inBackground) {
-        processingContext = CPUProcessingContext;
-    }
-    
-    CGSize maxInput = processingContext.inputImageMaximumSize;
-    CGSize maxOutput = processingContext.outputImageMaximumSize;
+    CGSize maxSize = CGSizeMake(1024, 1024);
```
appleguy: Is this a reasonable max size?  Perhaps so given that the blurring is supposed to be applied only to incompletely scanned progressive images - but what about a fullscreen iPad Pro at a near-final scan?  Perhaps arguably we shouldn't be blurring those scans at all, but I'm wondering if this value can be computed more abstractly.

Pod/Classes/PINProgressiveImage.m
```diff
@@ -350,59 +301,135 @@ - (CFTimeInterval)estimatedRemainingTime
...
+        maxSize.height < inputSize.height ||
+        maxSize.width < inputSize.width) {
         CGImageRelease(inputImageRef);
         return nil;
     }
     
-    if (processingContext && gaussianFilter) {
-        [gaussianFilter setValue:[CIImage imageWithCGImage:inputImageRef]
-                          forKey:kCIInputImageKey];
+    CGFloat radius = MAX(__FLT_EPSILON__, (inputImage.size.width / 25.0) * MAX(0, 1.0 - progress));
```
appleguy: Perhaps it's reasonable for this value to reach 0, and then just skip the blur if the calculation ever yielded that, instead of __FLT_EPSILON__?

Pod/Classes/PINProgressiveImage.m
```diff
@@ -350,59 +301,135 @@ - (CFTimeInterval)estimatedRemainingTime
...
-            
-            if (outputImageRef) {
-                outputUIImage = [UIImage imageWithCGImage:outputImageRef];
-                CGImageRelease(outputImageRef);
+        vImage_Buffer *inputBuffer;
+        vImage_Buffer *outputBuffer;
+        
+        vImage_CGImageFormat format = {
+            .bitsPerComponent = 8,
+            .bitsPerPixel = 32,
```
appleguy: Can we use an RGB - no alpha channel - buffer to save ~25% memory?  Or does vImage perhaps not support that with box convolution, as the function below does say "...8888" in the name?

appleguy:
@garrettmoon This looks great overall!  Very excited to test and profile!

Pod/Classes/PINProgressiveImage.m
garrettmoon: Good idea.

Pod/Classes/PINProgressiveImage.m
garrettmoon: Not that I can find, the box convolution functions operate on ARGB interleaved or planar 8 formatted images. We might be able to separate out the three channels to 3 separate planar 8 images and operate on them separately, but I'm not sure if that would result in the same image (I assume it would as my understanding of the box convolution just averages the surrounding pixels in each channel separately) or would have the same performance (though maybe you could calculate all three on separate threads).

But! In looking at this, I realize we don't need two image contexts and can get rid of one of them. So at least we have that going for us :)

PR #124, opened 2016-01-30
Move from dispatch_semaphore_t to XCTestExpectation

maicki:
Starting with Xcode 6, XCTest has been enhanced to include the ability to serialize asynchronous execution in the test method, by waiting for the completion of an asynchronous callback or timeout via XCTestExpectation. For most of test cases there is no need for dispatch_semaphore_t anymore.


garrettmoon:
Whoo! Thank you!

PR #125, opened 2016-01-31
Fix nullability warning in test

maicki:


garrettmoon:
Thanks for fixing this!

PR #126, opened 2016-02-02
Add generics to collections

rounak:
I've added generics to most places I could spot. Let me know if you spot any more places where it could be added.


garrettmoon:
Thanks!

PR #129, opened 2016-02-04
WebP + CocoaPods + use_frameworks! Redux

Adlai-Holler:
This supersedes #106. Tests pass locally and also I was able to use WebP in a Swift project with CocoaPods + use_frameworks!


Example/Podfile
```diff
@@ -1,3 +1,5 @@
+xcodeproj 'PINRemoteImage.xcodeproj'
```
Adlai-Holler: I initially built this with CocoaPods 0.39.0 and that version required this line. The current version for the project, 0.38.2, doesn't require it but also doesn't mind it so I left it. I can remove if needed.

Adlai-Holler:
Hate to intermingle pod sandbox updates with code changes like this, but it's the only way to get the tests running. The meaningful changes are all in d33db52.

Adlai-Holler:
OK so the remaining test failure is, on the iOS 8.4 simulator `testQOS` just timed out. I don't think it's related to this PR since the test didn't time out on 9.2 and I don't think I affected any of its tested code.

garrettmoon:
@Adlai-Holler the Travis CI builds seem to time out rather frequently, I restarted the failed one.

garrettmoon:
Thank you for doing this! I had to fix this up a bit for Mac support and the examples.

garrettmoon:
@Adlai-Holler I don't know why this says it's not merged, please confirm it is and then close.

Adlai-Holler:
Yep, looking good. Love this framework btw.

garrettmoon:
@Adlai-Holler Thank you for helping make it even better :)

PR #130, opened 2016-02-05
CoreImage is no longer used, instead we use accelerate

garrettmoon:


garrettmoon:
@maicki Can you confirm there were no specific CoreImage requirements for OS X?

PR #134, opened 2016-02-09
Add optional block to track download progress of image

maicki:
To get issue #70 going again I moved code from #110 over to a new pull request as well as updated it to the latest changes like nullability, OS X support etc.

Basically what it does is it adds an additional optional block to PINRemoteImageManager to allow users to track progress of image downloads in bytes.

@garrettmoon I didn't rename the progress parameter to progressImage yet as this would change the API. Don't know if it's already to late for 2.0 where maybe it would make sense to make API breaking changes ...

- [x] Add download progress image named: progressImage
- [x] Renamed progress to progressDownload
- [x] Call PINRemoteImageDownloadTask callback blocks outside the PINRemoteImageManager lock

### DIFF
--- a/Pod/Classes/PINRemoteImageManager.m
+++ b/Pod/Classes/PINRemoteImageManager.m
@@ -367,13 +367,13 @@ static dispatch_once_t sharedDispatchToken;
 {
     return [self downloadImageWithURL:url
                               options:options
-                             progress:nil
+                        progressImage:nil
                            completion:completion];
 }
 
 - (NSUUID *)downloadImageWithURL:(NSURL *)url
                          options:(PINRemoteImageManagerDownloadOptions)options
-                        progress:(PINRemoteImageManagerImageCompletion)progress
+                   progressImage:(PINRemoteImageManagerImageCompletion)progressImage
                       completion:(PINRemoteImageManagerImageCompletion)completion
 {
     return [self downloadImageWithURL:url
@@ -381,7 +381,7 @@ static dispatch_once_t sharedDispatchToken;
                              priority:PINRemoteImageManagerPriorityMedium
                          processorKey:nil
                             processor:nil
-                             progress:progress
+                        progressImage:progressImage
                      downloadProgress:nil
                            completion:completion
                             inputUUID:nil];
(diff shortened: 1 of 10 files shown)


garrettmoon:
@maicki if we can get the changes too, I think it'll probably be fine to move the 2.0 tag?

Pod/Classes/PINRemoteImageManager.m
```diff
@@ -1121,6 +1177,15 @@ - (void)didReceiveData:(NSData *)data forTask:(NSURLSessionDataTask *)dataTask
                 task.progressImage.progressThresholds = self.progressThresholds;
             }
         }
+    
+        for (PINRemoteImageCallbacks *completion in task.callbackBlocks.allValues) {
```
garrettmoon: Do you think it might be better to get a copy of the callbacks, and then do the actual callbacks outside of the lock? That way, blocking the callback won't lock up PINRemoteImage.

maicki: Yep agree let's change that

maicki:
@garrettmoon I will get that API change in ... just thinking you proposed to rename the "progress" parameter to "progressImage", but we have now a parameter that is called downloadProgress. I think that would not be consistent. What do you think rename the two either to "imageProgress" and "downloadProgress" or rename them to "progressImage" and "progressDownload"?

maicki:
I will get that API change in ... just thinking you proposed to rename the "progress" parameter to "progressImage", but we have now a parameter that is called "downloadProgress". I think that would be an inconstancy. What do you think about either calling them "imageProgress" and "downloadProgress" or  "progressImage" and "progressDownload"?

garrettmoon:
@maicki progressImage and progressDownload work for me!

maicki:
@garrettmoon Can you take another look over it please.

One thing maybe worth to look into too is the API around the progressive image e.g. setProgressThresholds:completion:. One way would be to rename it to setProgressiveImageThresholds:completion to make the distinction between the progress of an image and the progressive image a bit more clearer.

garrettmoon:
Thanks! Looks like Travis is having some issues, I'll test locally and merge.

PR #135, opened 2016-02-09
Replace UI* with PIN* category files in the PINRemoteImage project

maicki:
Seems like the category files in the PINRemoteImage project file wasn't updated from the UI* to the PIN* files as we added cross platform support. Currently the UI* category files show up as missing in the PINRemoteImage project. This pull request should fix that and replace the UI* category files with the PIN* category files.


garrettmoon:
Yiiissss, I have a similar branch, but it doesn't fix Carthage building. But this is on the path to getting things fixed up, thank you :D

PR #143, opened 2016-02-17
We only need to check respondsToSelector once and set QOS on queue, n…

garrettmoon:
…ot each operation.

### DIFF
--- a/Pod/Classes/PINRemoteImageManager.m
+++ b/Pod/Classes/PINRemoteImageManager.m
@@ -1431,7 +1431,7 @@ static dispatch_once_t sharedDispatchToken;
 {
     NSBlockOperation *operation = [NSBlockOperation blockOperationWithBlock:block];
     operation.queuePriority = operationPriorityWithImageManagerPriority(priority);
-    if ([[self class] supportsQOS] == NO) {
+    if ([PINRemoteImageManager supportsQOS] == NO) {
 #pragma clang diagnostic push
 #pragma clang diagnostic ignored "-Wdeprecated-declarations"
         operation.threadPriority = 0.2;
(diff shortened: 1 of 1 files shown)


Pod/Classes/PINRemoteImageManager.m
```diff
@@ -132,6 +132,15 @@ @implementation PINRemoteImageManager
 
 static PINRemoteImageManager *sharedImageManager = nil;
 static dispatch_once_t sharedDispatchToken;
+static BOOL supportsQOS = NO;
+
++ (void)load
+{
+    static dispatch_once_t onceToken;
+    dispatch_once(&onceToken, ^{
```
appleguy: Maybe better to do this in a method like +supportsQOS, and then just trigger it on first access rather than +load?

Pod/Classes/PINRemoteImageManager.m
```diff
@@ -1421,9 +1430,7 @@ - (void)pin_addOperationWithQueuePriority:(PINRemoteImageManagerPriority)priorit
 {
     NSBlockOperation *operation = [NSBlockOperation blockOperationWithBlock:block];
     operation.queuePriority = operationPriorityWithImageManagerPriority(priority);
-    if ([operation respondsToSelector:@selector(qualityOfService)]) {
-        operation.qualityOfService = NSOperationQualityOfServiceBackground;
-    } else {
+    if (supportsQOS == NO) {
 #pragma clang diagnostic push
 #pragma clang diagnostic ignored "-Wdeprecated-declarations"
         operation.threadPriority = 0.2;
```
appleguy: Sorta curious where this value came from.  Relevant for iPhone 4 / iOS 7, where this perhaps counts more than on any other device!

garrettmoon: It's absolutely a guess. If you have ideas for imperially figuring out the best value, I'd love to hear them!

PR #148, opened 2016-02-23
Fixes a crash in task callbacks

garrettmoon:
While all methods on tasks are called within the manager's lock,
dispatching to another queue obviously means the manager may not
be locked anymore.

I also moved calling progress download callbacks to the download
task so it's similar to progress image.

Hopefully these issues have been fixed, but it leads me to believe
a better architecture would be to make the tasks immutable by the
manager and have the tasks manage a lock on themselves. I'll need
to think about that more.


garrettmoon:
@maicki Mind taking a look at this PR, I moved some of your code.

rahul-malik:
@garrettmoon looks good to me!

Issue #151, opened 2016-02-24
Wrong define-based iOS/OSX determination


k06a:
Here is code to demonstrate this bug:
```
static int const aaa = __IPHONE_OS_VERSION_MIN_REQUIRED;
// Error: use of undeclared identifier '__IPHONE_OS_VERSION_MIN_REQUIRED'

#import <UIKit/UIKit.h>

static int const bbb = __IPHONE_OS_VERSION_MIN_REQUIRED;
// No error here
```

So this include will not work properly without importing UIKit:
```
#import <PINRemoteImage/PINRemoteImage.h>
```

And this code looks fun:
```
#ifdef __IPHONE_OS_VERSION_MIN_REQUIRED
#import <UIKit/UIKit.h>
#else
#import <Cocoa/Cocoa.h>
#endif
```

I'll try to describe this code literally: _if imported UIKit do import UIKit else import Cocoa_.

Maybe better to use this defines:
```
#if TARGET_OS_IPHONE || TARGET_OS_TV
#import <UIKit/UIKit.h>
#elif TARGET_OS_MAC
#import <Cocoa/Cocoa.h>
#endif
```

Found this solution here: https://github.com/SnapKit/Masonry/blob/master/Masonry/MASUtilities.h

garrettmoon:
Please put up a pull request!

k06a:
@garrettmoon no problem :) https://github.com/pinterest/PINRemoteImage/pull/155

garrettmoon:
@k06a Thanks again for putting up the PR, I think #164 fixed this.

PR #155, opened 2016-03-03
Fixed all OS checks

k06a:
Changes made in accordance to this issue: https://github.com/pinterest/PINRemoteImage/issues/151


garrettmoon:
Thanks for doing this @k06a ! I'll take a look soon.

garrettmoon:
I think this is addressed by #164 . Sorry for the churn!

k06a:
@garrettmoon anyway, the main thing that bug was fixed.

PR #157, opened 2016-03-06
Get XCPretty output

garrettmoon:


garrettmoon:
Need to remove testCancelDownload too because it's non-deterministic :( Cancelation isn't guaranteed.

PR #160, opened 2016-03-10
Fix Example-Mac and Carthage support

ikesyo:
Fixes https://github.com/Carthage/Carthage/issues/1179.

There is an invalid `Pods-PINRemoteImage-PINRemoteImage.xcscheme` scheme which does not have corresponding build target.


ikesyo:
Rebased.

garrettmoon:
Thanks @ikesyo !

PR #164, opened 2016-03-17
FIX: OS X + Carthage

justin:
This does two things

1. Make the target checks more explicit for tvOS & iOS vs OS X. 
2. Include the `TargetConditionals.h` file so that the `PINImage` macros play nicely when including a framework.

### DIFF
--- a/Cartfile.resolved
+++ b/Cartfile.resolved
@@ -1,1 +1,1 @@
-github "pinterest/PINCache" "2.2.1"
+github "pinterest/PINCache" "2.2.2"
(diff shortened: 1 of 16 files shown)


garrettmoon:
Do we need to check for WATCH_OS anywhere?

Separately, can you make an MACRO that expands to all the other targets? I.E.

`#define PIN_TARGET_IOS (TARGET_OS_IPHONE || TARGET_IPHONE_SIMULATOR || TARGET_OS_TV)

garrettmoon:
Thank you for doing this!

justin:
Are you wanting to support watchOS? :) I left it out because I didn't think it was a supported platform.

garrettmoon:
Ha, you're right, I was thinking PINCache.

garrettmoon:
Sorry to ask for one more change: Can you change the `@imports` back to `#imports`? `@import` doesn't work in Obj-C++ projects.

justin:
Done!

garrettmoon:
Thanks for doing this @justin!

PR #170, opened 2016-03-22
Add nullability specifier to PINURLSessionManager delegate

kerrmarin:
Hey,

I realised the delegate property was missing nullability specifiers, which can lead to compile-time errors on some configurations.

Cheers


garrettmoon:
Thanks!

PR #173, opened 2016-03-28
Add missing PINRemoteLock files to project.

dillan:
PINRemoteLock.[h|m] files were missing from the tvOS project, causing the build to fail. This would also cause build issues when attempting to include PINRemoteImage in a project using Carthage.

### DIFF
--- a/Example-tvOS/Pods/Pods.xcodeproj/project.pbxproj
+++ b/Example-tvOS/Pods/Pods.xcodeproj/project.pbxproj
@@ -46,6 +46,8 @@
 		A298D1FB9D743F0E70B1D0EA8F2523D1 /* PINRemoteImage-umbrella.h in Headers */ = {isa = PBXBuildFile; fileRef = F25810A52C69621B3D661F89EE87F20B /* PINRemoteImage-umbrella.h */; settings = {ATTRIBUTES = (Public, ); }; };
 		AAFF05A4CC3E4DFBB7E41B98EA30FB84 /* PINRemoteImage.h in Headers */ = {isa = PBXBuildFile; fileRef = 6BF71EE9F05CA5A8949C1EEA4E76F68E /* PINRemoteImage.h */; settings = {ATTRIBUTES = (Public, ); }; };
 		AB973F5515FA8617F9BF850FA22E136B /* PINCache.m in Sources */ = {isa = PBXBuildFile; fileRef = 8A34171F622CFB41A9D2950ABEA155CC /* PINCache.m */; };
+		B14CAE761CA8B82B00F552E0 /* PINRemoteLock.m in Sources */ = {isa = PBXBuildFile; fileRef = B14CAE741CA8B82B00F552E0 /* PINRemoteLock.m */; };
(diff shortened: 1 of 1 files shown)


garrettmoon:
I think the way to fix the tvOS project is to do a pod update in Example-tvOS directory.

dillan:
Does this mean that in order to use Carthage, Cocoa Pods must also be used?

garrettmoon:
@dillan no, the Example-tvOS needs CocoaPods though.

The carthage build is in https://github.com/pinterest/PINRemoteImage/tree/master/Example

That may need tvOS support separately?

dillan:
Interesting, when I try to include PINRemoteImage via Carthage the build fails without this change. 

I can understand an example project needing to use CocoaPods, but what I am after is being able use Carthage to build the frameworks. Right now it looks like Carthage is using the tvOS example project to build the framework target with tvOS symbols.

I've attached a sample project that exhibits this behavior when attempting to run:

`carthage update --use-submodules`

With that command I get the following output:

```
*** Fetching PINRemoteImage
*** Fetching PINCache
*** Checking out PINCache at "2.2.2"
*** Checking out PINRemoteImage at "2.1.2"
*** xcodebuild output can be found in /var/folders/l6/7zgxc03n0bl2ypw54lx_nkfh0000gn/T/carthage-xcodebuild.mrw8be.log
*** Building scheme "OS X" in PINCache.xcodeproj
*** Building scheme "iOS" in PINCache.xcodeproj
*** Building scheme "PINRemoteImage" in PINRemoteImage.tvOSExample.xcworkspace
** BUILD FAILED **

The following build commands failed:
	Ld /Users/dillan/Library/Developer/Xcode/DerivedData/PINRemoteImage.tvOSExample-bfojtipgtewupvbxkxfyvklzqtzz/Build/Products/Release-appletvos/PINRemoteImage.framework/PINRemoteImage normal arm64
(1 failure)
clang: error: linker command failed with exit code 1 (use -v to see invocation)
A shell task failed with exit code 65:
** BUILD FAILED **

The following build commands failed:
	Ld /Users/dillan/Library/Developer/Xcode/DerivedData/PINRemoteImage.tvOSExample-bfojtipgtewupvbxkxfyvklzqtzz/Build/Products/Release-appletvos/PINRemoteImage.framework/PINRemoteImage normal arm64
(1 failure)
```

[ImageTest.zip](https://github.com/pinterest/PINRemoteImage/files/194487/ImageTest.zip)

garrettmoon:
@dillan I think the PILock portion of the project has been updated, please let me know if not. I don't know why your Carthage is building the tvOSExample, that's weird, here's what I get:

```
*** Fetching PINCache
*** Checking out PINCache at "2.2.2"
*** xcodebuild output can be found in /var/folders/vz/bkb_rzmx3wd0h0gy91vdjs8r0000gn/T/carthage-xcodebuild.b12C16.log
*** Building scheme "OS X" in PINCache.xcodeproj
*** Building scheme "iOS" in PINCache.xcodeproj
```

It doesn't look like Carthage is set up to build tvOS at all?

I'm going to close this PR, feel free to open up an issue detailing what would help you if that helps (or ping me here if you think this change still needs to go in).

PR #175, opened 2016-03-29
Add support for custom decode

garrettmoon:
Adds support for custom decoding of images. Removes most FLAnimatedImage specific code, but retains support by default.


Example/PINRemoteImage Tests/PINRemoteImage_Tests.m
```diff
@@ -575,40 +575,21 @@ - (void)testProcessingLoad
...
     
     __block UIImage *image = nil;
     const NSUInteger numIntervals = 1000;
+    __block NSUInteger count = 0;
 
     for (NSUInteger idx = 0; idx < numIntervals; idx++) {
         dispatch_group_enter(group);
         [self.imageManager downloadImageWithURL:[self JPEGURL] completion:^(PINRemoteImageManagerResult *result) {
-            dispatch_group_leave(group);
+            count++;
```
timonus: `count` looks like it's unused here.

garrettmoon: Yes! Debugging, good catch!

Pod/Classes/PINAlternateRepresentationDelegate.h
```diff
@@ -0,0 +1,25 @@
...
+//
+//  Created by Garrett Moon on 3/17/16.
+//
+//
+
+#import <Foundation/Foundation.h>
+
+#import "PINRemoteImageManager.h"
+
+@protocol PINRemoteImageManagerAlternateRepresentationDelegate <NSObject>
```
rmls: It seems the only method here is providing an alternate representation, so I wonder if it would make sense to call this protocol `PINRemoteImageManagerAlternateRepresentationProvider` or something like it. Thoughts?

Pod/Classes/PINRemoteImageManager.m
```diff
@@ -1399,7 +1262,107 @@ - (NSUUID *)downloadImageWithURLs:(NSArray <NSURL *> *)urls
...
+        }
+        
+        container.data = data;
+    }
+    
+    if (alternateRepresentationsAllowed) {
+        alternateRepresentation = [_alternateRepDelegate alternateRepresentationWithData:data options:options];
+    }
+    
+    if (alternateRepresentation == NO) {
```
rmls: Shouldn't it be `alternateRepresentation != nil`?

Pod/Classes/PINAlternateRepresentationDelegate.h
```diff
@@ -0,0 +1,25 @@
...
+@required
+
+/**
+ @discussion This method will be called with data off the wire or stored in the cache. Return an object to have it returned as the alternativeRepresentation object in the PINRemoteImageManagerResult. @warning this method can be called on the main thread, be careful of doing expensive work.
+ */
+- (id)alternateRepresentationWithData:(NSData *)data options:(PINRemoteImageManagerDownloadOptions)options;
```
rmls: I assume you have considered this, but there's no way we can be more specific about the return of this method? Don't we need to conform to anything or be of anything? Can it literally be anything?

Pod/Classes/PINAlternateRepresentationDelegate.h
garrettmoon: That sounds better, will change.

Pod/Classes/PINRemoteImageManager.m
garrettmoon: should be alternateRepresentation == nil, good catch!

Pod/Classes/PINAlternateRepresentationDelegate.h
garrettmoon: I don't think so. The returned object isn't stored in the disk cache, so it doesn't need to conform to NSCoding. Otherwise, the only thing we do with this object is pass it through as a result, so it should be able to be an arbitrary object.

Pod/Classes/PINRemoteImageManager.h
```diff
@@ -127,13 +129,21 @@ typedef void(^PINRemoteImageManagerProgressDownload)(int64_t completedBytes, int
...
  */
 - (nonnull instancetype)initWithSessionConfiguration:(nullable NSURLSessionConfiguration *)configuration;
 
 /**
+ Create and return a PINRemoteImageManager with the specified configuration and alternative representation delegate. If configuration is nil, [NSURLSessionConfiguration defaultConfiguration] is used. Specify a custom configuration if you need to configure timeout values, cookie policies, additional HTTP headers, etc. If alternativeRepresentationDelegate is nil, the default is used (and supports FLAnimatedImage).
+ @param configuration The configuration used to create the PINRemoteImageManager.
+ @param alternativeRepDelegate a delegate which conforms to the PINRemoteImageManagerAlternateRepresentationDelegate protocol. Provide a delegate if you want to have non image results. @see PINAlternativeRepresentationDelaget for an example.
+ @return A PINRemoteImageManager with the specified configuration.
+ */
+- (nonnull instancetype)initWithSessionConfiguration:(nullable NSURLSessionConfiguration *)configuration alternativeRepresentationDelegate:(nullable id <PINRemoteImageManagerAlternateRepresentationDelegate>)alternateRepDelegate;
```
timonus: Might be nice to mark this `NS_DESIGNATED_INITIALIZER`. Probably not necessary since this isn't likely to be subclassed.

Pod/Classes/PINRemoteImageManager.m
```diff
@@ -100,6 +99,8 @@ @interface PINRemoteImageManager () <PINURLSessionManagerDelegate>
     PINRemoteLock *_lock;
     NSOperationQueue *_concurrentOperationQueue;
     NSOperationQueue *_urlSessionTaskQueue;
+    PINAlternateRepresentationDelegate *_defaultAlternateRepresentationDelegate;
```
timonus: Do we need this ivar? Seems like we could just have `_alternateRepDelegate` and assign it to a default in `init` without having a separate ivar just for the default delegate.

garrettmoon: Yes, _alternateRepDelegate is __weak. I should add a comment though.

garrettmoon:
Gonna address all y'all's comments in a follow-up PR.

Pod/Classes/PINRemoteImageManager.m
timonus: Oh duh, whoops. Good call.

timonus:
:+1:

PR #178, opened 2016-04-01
Added support for UIButton background images.

mad102190:
- Added a new category called PINRemoteBackgroundImageCategory to add support for UIButtons to download and set background images using PINRemoteImage. In order to set a background image on a view, the view must adhere to PINRemoteBackgroundImageCategory. 
- Added a launch image to the demo project to add support for running the sample app in the iPhone 6/6+ simulator in full screen.
- Added a new PINButtonCell class to the demo project to demonstrate the ability to download background images. 
- Added a switch to the main PINViewController class to allow switching from using PINImageCells and the new PINButtonCells


garrettmoon:
@mad102190 Instead of copying the implementations in PINRemoteImageCategoryManager.m, could you instead have all the existing methods take another parameter like an enum that indicates image type? For example you could have an enum that is PINViewTypeNormal, PINViewTypeBackground? What do you think?

mad102190:
Wouldn't this break existing implementations of the methods though? I initially thought of this, but after looking at similar libraries (i.e. SDWebImage), I figured it might be best to have specific methods for each image type.

garrettmoon:
@mad102190 I'm suggesting building it in a way that doesn't break existing implementations :)

Specifically adding all new category methods like you have, but instead of adding all new methods on PINRemoteImageCategoryManager, using the same implementation but adding an enum parameter which specifies the type?

mad102190:
Oh ok I see what you mean now. Sure, that works. I'll take a look shortly and update the PR.

garrettmoon:
@mad102190 I'm going to close this for now, if you decide to update it please reply and I can reopen.

PR #184, opened 2016-04-10
Modify estimatedRemainingTimeThreshhold Behavior to Match Documentation

Adlai-Holler:
- The documentation says the default is 0.1 but it actually defaults to 0. Updated the documentation.
- The documentation says if you set time <= 0 the check is skipped, but actually if you set time < 0 progressive images will never be rendered. Updated the behavior.
- The documentation says maxProgressiveRenderSize only affects which images are blurred, but it actually affects which images are progressively rendered at all. Updated the documentation.

### DIFF
--- a/Pod/Classes/PINProgressiveImage.m
+++ b/Pod/Classes/PINProgressiveImage.m
@@ -165,7 +165,7 @@
             return nil;
         }
         
-        if (_estimatedRemainingTimeThreshold < 0 || self.estimatedRemainingTime < _estimatedRemainingTimeThreshold) {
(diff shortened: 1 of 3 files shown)


garrettmoon:
Good catches! Thank you!

I think that the estimatedRemainingTimeThreshold *should* default to 0.1. Mind making that change instead of the change to the documentation?

Adlai-Holler:
@garrettmoon Sure man! Done and done.

garrettmoon:
Thanks!

PR #189, opened 2016-04-18
Use a hashed key when the URL is too long

dtorres:
The filesystem doesn't support names longer than 255 chars, when that happens the file is not saved, rendering the disk cache useless.
 
This fixes that by creating a hashed key in those cases.

### DIFF
--- a/Pod/Classes/PINRemoteImageManager.m
+++ b/Pod/Classes/PINRemoteImageManager.m
@@ -9,6 +9,7 @@
 #import "PINRemoteImageManager.h"
 
 #import <PINCache/PINCache.h>
+#import <CommonCrypto/CommonDigest.h>
 
 #import "PINAlternateRepresentationProvider.h"
 #import "PINRemoteImage.h"
@@ -26,6 +27,10 @@
 #import "PINImage+DecodedImage.h"
 
 #define PINRemoteImageManagerDefaultTimeout  60.0
+
+//A limit of 200 characters is chosen because PINDiskCache
+//may expand the length by encoding certain characters
+#define PINRemoteImageManagerCacheKeyMaxLength 200
 
 NSOperationQueuePriority operationPriorityWithImageManagerPriority(PINRemoteImageManagerPriority priority) {
     switch (priority) {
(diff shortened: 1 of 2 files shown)


garrettmoon:
I think the issue should be addressed in PINCache. Or at the very least, PINCache should throw an exception if you try and use a key that is too long.

Separately, cacheKeyForURL is called on the main thread, so I'd be concerned that this could cause performance issues.

dtorres:
Agreed that PINDiskCache should throw an exception.

The problem with addressing it in PINCache (And that's the first I looked to fix it in) is that it means to change the keys internally to some canonical form, and then use that form in both the memory and disk cache.

Also, PINDiskCache uses the key as the file name and even performs a "decoding" of the path to get the key, so it is not possible to internally address the issue in that class without changing the key and therefore going up the chain.

To enforce a key length limit needlessly restricts the usage of the API when a potential user might not need the disk cache.

Lastly, this issue is directly related to URLs, hence why I think the best address it completely at this level.

About the performance concerns:
At any level that we solve this, key Creation is synchronous so we will eventually hit the main thread anyways.

I suggest adding a NSCache to perform the key calculation the least times possible. And also, as you might see, it only hits keys which are very large and likely to not be working anyways.

Thoughts?

garrettmoon:
My biggest concern is calculating a SHA on the main thread. You're right that it will only occur in cases where the items would be dropped from the disk cache, but we'll be blocking to calculate the SHA even when the item is in the memory cache.

If there's no way to reduce the complexity, an NSCache might mitigate the performance effects, but I'd rather find a way to reduce the performance effects across the board.

SHA is designed for cryptographic use. There must be better hashing algorithms that aren't optimized for cryptography?

dtorres:
True that...
two algorithms come to mind, MurmurHash3 and MD5.

Murmur is slightly faster (http://blog.httrack.com/blog/2013/07/19/md5-is-your-friend/) in hashing small keys (as in URLs) but MD5 is readily available in CommonCrypto and means not having another dependency. The difference is negligible.

garrettmoon:
You present a compelling argument, MD5 is probably reasonable and we can improve it in the future if it comes up.

Can you switch it to MD5 and add unit tests? I've opened an issue on PINCache: https://github.com/pinterest/PINCache/issues/80

garrettmoon:
Oh! You'll also need to add common crypto as a dependency in the podfile

dtorres:
Will do, tomorrow. 

About the dependency, I meant an external one, and you are right, forgot to commit that!

garrettmoon:
I understood about the dependency :)

dtorres:
CommonCrypto is not a framework and as such apparently it is included in the default loaded libraries. The tests run fine without a change in the podspec. I don't know how to trigger the CI again but it says it failed because I previously pushed a change in the podspec which broke the linting

Pod/Classes/PINRemoteImageManager.m
```diff
@@ -1392,6 +1397,29 @@ - (NSString *)cacheKeyForURL:(NSURL *)url processorKey:(NSString *)processorKey
...
+        }];
+
+        unsigned char digest[CC_MD5_DIGEST_LENGTH];
+        CC_MD5_Final(digest, &ctx);
+
+        NSMutableString *hexString  = [NSMutableString stringWithCapacity:(CC_MD5_DIGEST_LENGTH * 2)];
+        for (int i = 0; i < CC_MD5_DIGEST_LENGTH; i++) {
+            [hexString appendFormat:@"%02lx", (unsigned long)digest[i]];
+        }
+        cacheKey = hexString.copy;
```
garrettmoon: mind making this cacheKey = [hexString copy]; to match the styling in the rest of the project?

garrettmoon:
Yeah, I don't know what's up with the tests. I'll get them figured out.

Pod/Classes/PINRemoteImageManager.m
dtorres: Sure, as soon as I'm in front of my computer

PR #191, opened 2016-04-22
Generate the cacheKey only when necessary

dtorres:
This addresses performance concerns raised in #189 by reducing calls to the cacheKey generation.


garrettmoon:
I like this better beyond the performance implications! Thanks!

PR #199, opened 2016-05-07
[PINProgressiveImage] Use CoreImage Rather than vImage for Blurring

Adlai-Holler:
Since this uses the GPU directly, the performance running on iOS Simulator is very bad.

But on my iPhone 6, this performed more consistently and slightly faster in the example app than the prior Accelerate-based approach. Under accelerate, I saw between 125-600 ms per progressive render. With this approach, the range was 120-240ms. 

I haven't tested this under OS X yet. Is there an OS X project in the repo?

### DIFF
--- a/Pod/Classes/PINProgressiveImage.m
+++ b/Pod/Classes/PINProgressiveImage.m
@@ -325,25 +325,25 @@
         return inputImage;
     }
 
-	// Creating CI contexts is expensive. We store them in a reuse pool to improve performance.
+    // Creating CI contexts is expensive. We store them in a reuse pool to improve performance.
 
-	// TODO: Is it worth clearing this pool on memory warning? Probably not, since we're only going
-	// to create as many contexts as we have concurrent blur operations, but we need to profile
-	// the memory consumption of CIContext.
-	static NSMutableArray *contexts;
-	static NSLock *contextsLock;
-	static dispatch_once_t onceToken;
-	dispatch_once(&onceToken, ^{
+    // TODO: Is it worth clearing this pool on memory warning? Probably not, since we're only going
+    // to create as many contexts as we have concurrent blur operations, but we need to profile
+    // the memory consumption of CIContext.
+    static NSMutableArray *contexts;
+    static NSLock *contextsLock;
+    static dispatch_once_t onceToken;
+    dispatch_once(&onceToken, ^{
         contexts = [NSMutableArray array];
-		contextsLock = [[NSLock alloc] init];
-	});
+        contextsLock = [[NSLock alloc] init];
+    });
     
-	[contextsLock lock];
+    [contextsLock lock];
         CIContext *context = [contexts lastObject];
-		if (context != nil) {
+        if (context != nil) {
             [contexts removeLastObject];
-		}
-	[contextsLock unlock];
+        }
+    [contextsLock unlock];
 
     if (context == nil) {
         // NOTE: We use the software renderer because accessing the GPU when the app
@@ -390,10 +390,10 @@
 
     CGImageRef outputImageRef = [context createCGImage:blurred fromRect:bounds];
 
-	// Return our context to the reuse pool.
-	[contextsLock lock];
-		[contexts addObject:context];
-	[contextsLock unlock];
+    // Return our context to the reuse pool.
+    [contextsLock lock];
+        [contexts addObject:context];
+    [contextsLock unlock];
 
 #if PIN_TARGET_IOS
     outputImage = [UIImage imageWithCGImage:outputImageRef];
--- a/Pod/Classes/PINProgressiveImage.m
+++ b/Pod/Classes/PINProgressiveImage.m
@@ -228,7 +228,7 @@
             CGImageRef image = CGImageSourceCreateImageAtIndex(self.imageSource, 0, NULL);
             if (image) {
                 if (blurred) {
-                    currentImage = [self postProcessImage:[PINImage imageWithCGImage:image] withProgress:progress];
+                    currentImage = [PINProgressiveImage postProcessImage:[PINImage imageWithCGImage:image] withProgress:progress];
                 } else {
                     currentImage = [PINImage imageWithCGImage:image];
                 }
@@ -311,8 +311,7 @@
     return remainingBytes / self.bytesPerSecond;
 }
 
-//Must be called within lock
-- (PINImage *)postProcessImage:(PINImage *)inputImage withProgress:(float)progress
++ (PINImage *)postProcessImage:(PINImage *)inputImage withProgress:(float)progress
 {
     PINImage *outputImage = nil;
     CIImage *inputCIImage = [CIImage imageWithCGImage:inputImage.CGImage options:nil];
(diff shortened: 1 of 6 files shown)


Adlai-Holler:
Also any thoughts about software rendering here? I don't want to waste memory bandwidth transferring back and forth from the GPU but at the same time convolution is very parallelizable. In the sandbox app the performance difference was negligible but I'm thinking about cases where you've got a feed full of progressive images. Can test but if you have some insight here that'd be helpful cc @appleguy

garrettmoon:
@Adlai-Holler So this is actually what PINRemoteImage used to do :(

I pulled it out because it was the source of crashes and I figured the CPU cost (off the main thread) was worth it. I'd be open to putting this implementation back in if we could somehow verify it's not crashing anymore, but I spent a *long* time trying to address the crashes and came to the conclusion they might be in CI :/

garrettmoon:
@Adlai-Holler And to answer your question, yes there's an OS X project at `/Example-Mac`

garrettmoon:
Also, I really appreciate you putting this up, I'm sad it may be wasted work :(

garrettmoon:
@Adlai-Holler if I'm not mistaken, using the GPU in the background will cause CoreImage to crash the program. The old CI implementation used notifications to switch to the software renderer if the app was backgrounded. One of the other advantages to switching to accelerate was making sure PINRemoteImage didn't rely on UIApplication notifications (and thus needing separate code paths for the extension).

Adlai-Holler:
Sorry I missed your previous replies. Well it's no good if it's going to crash things! I threw my profiler results away but I remember that using CoreImage + software rendering was faster than accelerate so let me do these things, then I'll come back here and update:

- Ensure that using CoreImage software rendering with the app in the background doesn't cause problems and is allowed by the API contract
- Benchmark CoreImage+software vs accelerate again so we can discuss the benefits objectively.

garrettmoon:
Sounds good! I definitely *want* Core Image to work :)

Adlai-Holler:
Hey sorry for the delay here – I got sick the next morning and I'm just now recovering. Haven't forgotten!

appleguy:
@Adlai-Holler just wanted to support what @garrettmoon said here - we both would like this to work, but there are a few concerns.

- Using OpenGL / GPU in the background is most definitely forbidden and causes intentional crashes (abort called internal to Apple frameworks) when it occurs.  This is fairly difficult to prevent as we have some indication that even operations in progress, e.g. large images, may hit this case.
- Various versions of iOS, on specific types of hardware, have GPU drivers with bugs that result in crashes due to large numbers of context switches that result from heavily parallelized use of OpenGL - because the GPU is inherently single-threaded (despite parallel units internally), there is a significant thrash as the render server is trying to composite frames and grabbing the GPU to work on an image blur.
- Especially on older devices, which are the same ones where this difference is most relevant, taking the GPU from the render server actually results in significant increases in frame drops that can be difficult to measure.  This is because devices like iPhone 4, iPad 3, and even 4S, iPad 2, 4, mini 1, etc are all commonly at or near being GPU-bound.  As far as I know, the GPU can't context switch out of an operation in progress, so although blurring is much faster in terms of ms than on the CPU, it can still lead to small or potentially large frame drops.  There is also setup cost to the OpenGL context and the GPU context switch (including the memory bandwidth point).

Overall, it is fairly difficult to analyze these subtleties.  I spent a lot of time looking at this way back in 2011 when some special Apple tools were available to me.  Although hardware has changed a lot, some of the devices of that day (or comparable) are still in wide use (especially iPad 2-4 and mini 1 which make up ~50% of active devices).

My suggestion is to come check out some of our Pinterest crash logs in the PowerVR SGX driver, and weep with us over a fine scotch.

Adlai-Holler:
Hahaha ok I won't poke that particular bear for now. Do you know if using the software rendering option suffers from any of those issues?

garrettmoon:
@Adlai-Holler I can't say for certain, but the old code would switch to a software renderer when the app was backgrounded. It's entirely possible that this always happened too late, but I seem to remember that the crash rate didn't go down when we were more aggressive about using the software renderer. That said, I'd be happy to look at some of the crash reports together :)

Adlai-Holler:
OK! Some interesting developments here that I'm happy to report! Software CoreImage is delivering in a big way!

Creating CIContexts turns out to be very expensive. I've implemented a simple reuse pool of them with an NSMutableSet and NSLock. If we decide to go with this code, I can refine this implementation (maybe break it out into a singleton object that listens for memory warnings etc.)

With the reuse pool in place, CoreImage outperforms the Accelerate implementation very significantly. I've attach the raw test results below. Also, note that after excluding the first test case, the gap widens even further, avg 0.117 vs avg 0.364.

I still need to confirm that this won't cause crashes. Obviously this is tough to get reliable data on, but my intention is just to play with Tripstr (lots of progressive JPEG) over and over again on 3G. 

Raw test results on my iPod touch 5. Each subcase (each result) is 10 iterations on a concurrent queue blurring a medium JPEG with 0.5 progress.

```
Test Case '-[PINRemoteImage_Tests testImageBlurringPerformance_CoreImage]' measured [Time, seconds] average: 0.167, relative standard deviation: 90.374%, values: [0.615826, 0.107276, 0.142145, 0.103781, 0.099211, 0.113100, 0.163194, 0.102095, 0.108695, 0.113901]

Test Case '-[PINRemoteImage_Tests testImageBlurringPerformance_Accelerate]' measured [Time, seconds] average: 0.371, relative standard deviation: 6.127%, values: [0.430207, 0.355882, 0.360598, 0.355662, 0.362676, 0.394954, 0.369209, 0.362731, 0.356487, 0.358493]
```

Thoughts @garrettmoon @appleguy ?

Pod/Classes/PINProgressiveImage.m
```diff
@@ -311,163 +311,95 @@ - (CFTimeInterval)estimatedRemainingTime
...
+	});
+	[contextsLock lock];
+		CIContext *context = [contexts anyObject];
+		if (context != nil) {
+			[contexts removeObject:context];
+		} else {
+			// NOTE: We use the software renderer because accessing the GPU when the app
+			// is not in the foreground is unsafe. We have attempted switching to the software
+			// renderer when the app goes into the background but crashes still remain.
+			context = [CIContext contextWithOptions:@{ kCIContextUseSoftwareRenderer: @YES }];
```
jparise: I think you can move new context creation outside of the lock:

```objc
[contextsLock lock];
    CIContext *context = [contexts anyObject];
    if (context != nil) {
        [contexts removeObject:context];
    }
[contextsLock unlock];

if (context == nil) {
    context = [CIContext contextWithOptions:@{ kCIContextUseSoftwareRenderer: @YES }];
}
```

garrettmoon:
One thing to note about the context pool: the documentation suggests that CIContexts themselves are immutable and thread safe. So you should be able to just use one. That said, maybe that was the source of our crashes… (we had one context for the software renderer and one for the GPU renderer if I remember correctly, old implementation should be in git history).

Pod/Classes/PINProgressiveImage.m
Adlai-Holler: @jparise Ah good catch! Change made.

Adlai-Holler:
@garrettmoon Ah you're right! I've set up a macro to switch between the reuse pool and the shared context, currently set to shared context. We'll see what we can get away with there!

rahul-malik:
@Adlai-Holler @appleguy @garrettmoon - Any resolution on this PR? Are there investigations still happening or should we abandon this change?

appleguy:
@rahul-malik IMO we should abandon.  It can be revived if needed.  However, due to known thread safety issues with CI and the opacity of its internals, I think we will use other approaches.

garrettmoon:
Closing for now.

PR #205, opened 2016-06-07
Fix runaway NSData

garrettmoon:
Previously we were generating all the frames of the GIF and dispatching
to write the data to the disk. This presents a problem if the disk backs
up though and it can end up with theoretically every frame in memory. Instead,
generate a frame and write it one frame at a time. Slower, but prevents
this runaway memory situation.

### DIFF
--- a/Pod/Classes/PINAnimatedImageManager.m
+++ b/Pod/Classes/PINAnimatedImageManager.m
@@ -380,34 +380,40 @@ ERROR;}) \
               });
             }
             
-            CGImageRef frameImage = CGImageSourceCreateImageAtIndex(imageSource, frameIdx, (CFDictionaryRef)@{(__bridge NSString *)kCGImageSourceShouldCache : (__bridge NSNumber *)kCFBooleanFalse});
-            if (frameImage == nil) {
-              NSError *frameImageError = [NSError errorWithDomain:kPINAnimatedImageErrorDomain code:PINAnimatedImageErrorImageFrameError userInfo:nil];
-              HANDLE_PROCESSING_ERROR(frameImageError);
-              break;
-            }
-            
             Float32 duration = durations[frameIdx];
             fileDuration += duration;
-            NSData *frameData = (__bridge_transfer NSData *)CGDataProviderCopyData(CGImageGetDataProvider(frameImage));
-            NSAssert(frameData.length == width * height * kPINAnimatedImageComponentsPerPixel, @"data should be width * height * 4 bytes");
+            
             dispatch_group_async(diskGroup, diskWriteQueue, ^{
+              if (PROCESSING_ERROR) {
+                return;
+              }
+              
+              CGImageRef frameImage = CGImageSourceCreateImageAtIndex(imageSource, frameIdx, (CFDictionaryRef)@{(__bridge NSString *)kCGImageSourceShouldCache : (__bridge NSNumber *)kCFBooleanFalse});
+              if (frameImage == nil) {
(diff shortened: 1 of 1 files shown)


garrettmoon:
Before is on bottom memory usage when decoding 3 GIFs, after is on top.

screen shot 2016-06-07 at 2 32 04 pm

Pod/Classes/PINAnimatedImageManager.m
```diff
@@ -380,34 +380,40 @@ + (void)processAnimatedImage:(NSData *)animatedImageData
...
+                NSError *frameImageError = [NSError errorWithDomain:kPINAnimatedImageErrorDomain code:PINAnimatedImageErrorImageFrameError userInfo:nil];
+                HANDLE_PROCESSING_ERROR(frameImageError);
+              }
+              
+              NSData *frameData = (__bridge_transfer NSData *)CGDataProviderCopyData(CGImageGetDataProvider(frameImage));
+              NSAssert(frameData.length == width * height * kPINAnimatedImageComponentsPerPixel, @"data should be width * height * 4 bytes");
               NSError *frameWriteError = [self writeFrameToFile:fileHandle duration:duration frameData:frameData];
               HANDLE_PROCESSING_ERROR(frameWriteError);
+              
+              CGImageRelease(frameImage);
```
rahul-malik: Do we need to protect this call if the frameImage is nil?

Pod/Classes/PINAnimatedImageManager.m
```diff
@@ -380,34 +380,40 @@ + (void)processAnimatedImage:(NSData *)animatedImageData
...
+            
             dispatch_group_async(diskGroup, diskWriteQueue, ^{
+              if (PROCESSING_ERROR) {
+                return;
+              }
+              
+              CGImageRef frameImage = CGImageSourceCreateImageAtIndex(imageSource, frameIdx, (CFDictionaryRef)@{(__bridge NSString *)kCGImageSourceShouldCache : (__bridge NSNumber *)kCFBooleanFalse});
+              if (frameImage == nil) {
+                NSError *frameImageError = [NSError errorWithDomain:kPINAnimatedImageErrorDomain code:PINAnimatedImageErrorImageFrameError userInfo:nil];
+                HANDLE_PROCESSING_ERROR(frameImageError);
```
rahul-malik: should we early return here?

Pod/Classes/PINAnimatedImageManager.m
garrettmoon: CGImageRelease is basically CFRelease with a nil check (according to docs).

Pod/Classes/PINAnimatedImageManager.m
```diff
@@ -380,34 +380,40 @@ + (void)processAnimatedImage:(NSData *)animatedImageData
...
+              }
+              
+              NSData *frameData = (__bridge_transfer NSData *)CGDataProviderCopyData(CGImageGetDataProvider(frameImage));
+              NSAssert(frameData.length == width * height * kPINAnimatedImageComponentsPerPixel, @"data should be width * height * 4 bytes");
               NSError *frameWriteError = [self writeFrameToFile:fileHandle duration:duration frameData:frameData];
               HANDLE_PROCESSING_ERROR(frameWriteError);
```
timonus: Similar to what @rahul-malik mentioned, should we only execute this if `frameImage` is non-`nil`?

rahul-malik:
Looks good to me after the updates

PR #219, opened 2016-07-15
Without this, the system is likely caching our NSURLResponses which m…

garrettmoon:
…eans double caching :(


jparise:
👍

maicki:
@garrettmoon You have to be careful it was introduces in OS X 10.9 and we have in the Podspec we also support OS X 10.8

garrettmoon:
@maicki Thanks! I'll check that.

garrettmoon:
@maicki Oh! NSURLSession is unavailable in 10.8. We'll need to update the pod spec :/

Issue #169, opened 2016-03-21
Consider calling completion and progress callbacks on the main thread


DavidNix:
Line of interest:
https://github.com/pinterest/PINRemoteImage/blob/7ba1bee59c733f4e11267700e0d4a51f5a177404/Pod/Classes/PINRemoteImageManager.m#L176

It's very common for progress and completion callbacks to update UI elements.  Hence, as a library author, calling them on the main thread is typical.  I would even argue best practice.  I admit, I haven't had the time to fully understand the need for calling them on an arbitrary queue (instead of the main queue).  The documentation hints at optimization reasons.

I spent a good while banging my head because these callbacks were not updating the UI:

```
// From PINRemoteImageManager.h
- (nullable NSUUID *)downloadImageWithURL:(nonnull NSURL *)url
                                  options:(PINRemoteImageManagerDownloadOptions)options
                         progressDownload:(nullable PINRemoteImageManagerProgressDownload)progressDownload
                               completion:(nullable PINRemoteImageManagerImageCompletion)completion;
```

Then I read the documentation more closely (my fault) and noticed:

> Download or retrieve from cache the image found at the url. All completions are called on an arbitrary callback queue unless called on the main thread and the result is in the memory cache (this is an optimization to allow synchronous results for the UI when an object is cached in memory).

I was using the progress block to update a progress bar in the UI as well as the completion block to set an image on a UIImageView.

Perhaps a config option to force callbacks on the main queue?

Thanks for this great library.

garrettmoon:
@DavidNix Thanks for the feedback!

Yeah, this is definitely something that snags a lot of devs. It's part of the reason all the default categories exist.

That said, there are very specific performance reasons to not use the main thread. Maybe a clearer warning in the README would help?

DavidNix:
The reason I dipped into the Manager is to get at that progress callback.
Perhaps expose the progress block in a category method?

But a big warning in the README would definitely help.

Thanks!

On Mon, Mar 21, 2016 at 4:04 PM Garrett Moon <notifications@github.com>
wrote:

> @DavidNix <https://github.com/DavidNix> Thanks for the feedback!
>
> Yeah, this is definitely something that snags a lot of devs. It's part of
> the reason all the default categories exist.
>
> That said, there are very specific performance reasons to not use the main
> thread. Maybe a clearer warning in the README would help?
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly or view it on GitHub
[...]

PR #222, opened 2016-07-16
Configurable callback queue #169

danielgalasko:
This PR exposes the callback queue on PINRemoteImageManager and having it default to the main thread.

This makes it less likely for this library to trip up most developers and now makes it so that those comfortable with async dev can set a different queue.

This was discussed in #169 

## TODO:

- [ ] Add initialiser that accepts a callback queue rather than allow it to be set at any point in the managers lifetime

### DIFF
--- a/Example/PINRemoteImage Tests/PINRemoteImage_Tests.m
+++ b/Example/PINRemoteImage Tests/PINRemoteImage_Tests.m
@@ -234,6 +234,37 @@
     [self waitForExpectationsWithTimeout:[self timeoutTimeInterval] handler:nil];
 }
 
+- (void)testCallbackQueueDefaultsToMainQueue
+{
+    dispatch_queue_t callbackQueue = self.imageManager.callbackQueue;
+    dispatch_queue_t mainqueue = dispatch_get_main_queue();
+    XCTAssert(callbackQueue == mainqueue);
+    XCTestExpectation *expectation = [self expectationWithDescription:@"Downloading JPEG image"];
+    [self.imageManager downloadImageWithURL:[self JPEGURL]
+                                    options:PINRemoteImageManagerDownloadOptionsNone
+                                 completion:^(PINRemoteImageManagerResult *result)
+     {
+         XCTAssertTrue([[NSThread currentThread] isMainThread]);
+         [expectation fulfill];
+     }];
+    [self waitForExpectationsWithTimeout:[self timeoutTimeInterval] handler:nil];
+}
+
+- (void)testCanUseDifferentCallbackQueue {
+    dispatch_queue_t callbackQueue = dispatch_queue_create("PINRemoteImageManagerCallbackQueue", DISPATCH_QUEUE_CONCURRENT);
+    self.imageManager.callbackQueue = callbackQueue;
+    XCTestExpectation *expectation = [self expectationWithDescription:@"Downloading JPEG image"];
+    [self.imageManager downloadImageWithURL:[self JPEGURL]
+                                    options:PINRemoteImageManagerDownloadOptionsNone
+                                 completion:^(PINRemoteImageManagerResult *result)
+     {
+         XCTAssertFalse([[NSThread currentThread] isMainThread]);
+         [expectation fulfill];
+     }];
+    [self waitForExpectationsWithTimeout:[self timeoutTimeInterval] handler:nil];
+    
+}
+
 - (void)testErrorOnNilURLDownload
 {
 #pragma clang diagnostic push
--- a/Example/PINRemoteImage Tests/PINRemoteImage_Tests.m
+++ b/Example/PINRemoteImage Tests/PINRemoteImage_Tests.m
@@ -258,7 +258,7 @@
                                     options:PINRemoteImageManagerDownloadOptionsNone
                                  completion:^(PINRemoteImageManagerResult *result)
      {
-         XCTAssertFalse([[NSThread currentThread] isMainThread]);
+         XCTAssertFalse([NSThread isMainThread]);
          [expectation fulfill];
      }];
     [self waitForExpectationsWithTimeout:[self timeoutTimeInterval] handler:nil];
--- a/Example/PINRemoteImage Tests/PINRemoteImage_Tests.m
+++ b/Example/PINRemoteImage Tests/PINRemoteImage_Tests.m
@@ -244,7 +244,7 @@
                                     options:PINRemoteImageManagerDownloadOptionsNone
                                  completion:^(PINRemoteImageManagerResult *result)
      {
-         XCTAssertTrue([[NSThread currentThread] isMainThread]);
+         XCTAssertTrue([NSThread isMainThread]);
          [expectation fulfill];
      }];
     [self waitForExpectationsWithTimeout:[self timeoutTimeInterval] handler:nil];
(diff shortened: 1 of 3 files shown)


danielgalasko:
Looks like travis is bust?

garrettmoon:
Travis often lies :( I'll take a look at this as soon as I'm able, hopefully before the end of the week.

Example/PINRemoteImage Tests/PINRemoteImage_Tests.m
```diff
@@ -234,6 +234,37 @@ - (void)testJPEGDownload
...
+
+- (void)testCanUseDifferentCallbackQueue {
+    dispatch_queue_t callbackQueue = dispatch_queue_create("PINRemoteImageManagerCallbackQueue", DISPATCH_QUEUE_CONCURRENT);
+    self.imageManager.callbackQueue = callbackQueue;
+    XCTestExpectation *expectation = [self expectationWithDescription:@"Downloading JPEG image"];
+    [self.imageManager downloadImageWithURL:[self JPEGURL]
+                                    options:PINRemoteImageManagerDownloadOptionsNone
+                                 completion:^(PINRemoteImageManagerResult *result)
+     {
+         XCTAssertFalse([[NSThread currentThread] isMainThread]);
```
rahul-malik: nit: It's shorter to just use the class method version:  `[NSThread isMainThread]`

Example/PINRemoteImage Tests/PINRemoteImage_Tests.m
```diff
@@ -234,6 +234,37 @@ - (void)testJPEGDownload
...
+                                 completion:^(PINRemoteImageManagerResult *result)
+     {
+         XCTAssertTrue([[NSThread currentThread] isMainThread]);
+         [expectation fulfill];
+     }];
+    [self waitForExpectationsWithTimeout:[self timeoutTimeInterval] handler:nil];
+}
+
+- (void)testCanUseDifferentCallbackQueue {
+    dispatch_queue_t callbackQueue = dispatch_queue_create("PINRemoteImageManagerCallbackQueue", DISPATCH_QUEUE_CONCURRENT);
```
rahul-malik: Is there a particular reason to supply a concurrent queue or is that simply just to differentiate it from the main queue?

Pod/Classes/PINRemoteImageManager.m
```diff
@@ -189,7 +188,7 @@ - (instancetype)initWithSessionConfiguration:(NSURLSessionConfiguration *)config
         if (!configuration) {
             configuration = [NSURLSessionConfiguration ephemeralSessionConfiguration];
         }
-        _callbackQueue = dispatch_queue_create("PINRemoteImageManagerCallbackQueue", DISPATCH_QUEUE_CONCURRENT);
+        _callbackQueue = dispatch_get_main_queue();
```
rahul-malik: Since this queue can be changed externally at any time, we will either need to protect the getter / setters of this properties with a PINRemoteLock instance for thread-safety. Another alternative is to create another initializer where the caller can specify the queue and remove the property from the public interface entirely.

danielgalasko: I actually thought about this and then looked at how AFNetworking did this and I don't think supporting modification of a live queue makes sense (they just expose it as a nonatomic property). This is something that should only be set once. Or do you think there is a good reason to add extra support for this?

Example/PINRemoteImage Tests/PINRemoteImage_Tests.m
danielgalasko: You mean as another test case? Definitely because it defaults to the main queue so its nice to see if we can change this. or am I missing the question?

Example/PINRemoteImage Tests/PINRemoteImage_Tests.m
danielgalasko: Yes! will update

garrettmoon:
@danielgalasko Every other property on PINRemoteImage is thread safe. So this var must be accessible through the init method or have thread safe getters and setters.

Separately, I'm not excited about making the callback queue default to the main queue. All the PINRemoteImage categories handle dispatching to the main thread and most clients should use those in UI work. If they choose to use the download methods directly, the documentation clearly states that the callbacks will happen on an arbitrary queue.

**If this change is to go in, all the documentation will need to be updated as well.**

Pod/Classes/PINRemoteImageManager.m
rahul-malik: I think this should only be set once for now through the initializer. If there is a good reason to swap the callback queue at runtime we can revisit but in the meanwhile this will satisfy the thread-safety concerns here.

Example/PINRemoteImage Tests/PINRemoteImage_Tests.m
rahul-malik: No particular concern here - Just asking if there was a reason to specify a concurrent vs serial queue to supply to the test here.

danielgalasko:
@garrettmoon This is actually something I have been thinking about for a while. Given how many times I have seen new iOS devs get tripped up by using UIKit on a background thread because the network layer doesn't dispatch to main, I keep questioning the reason why we always use background threads for callbacks given that 99.99% of the time the image result will be needed by the main queue.

So IMO its the more advanced developers who will be setting the callback queue to a background queue if they want to opt in to image processing or some other funky stuff. 

But I would love to discuss this and hear your thoughts on the merits of defaulting to the background thread. I am also coming from a perspective of always preferring background threads but recent conversations I have had with fellow devs have made me question this. My thinking is to pretty much make it as fool-proof as possible for newbies and have the 1% of use cases that need a background queue do their own dispatching, or if they want all calls backgrounded they can set the callBack queue. This way code doesn't get littered with dispatches to main.

Thoughts?

Pod/Classes/PINRemoteImageManager.m
danielgalasko: Fair enough, will do this now

garrettmoon:
I feel we've likely brought up most of the arguments regarding main vs. non-main queue. My opinion is that we're doing new developers no favors by protecting them. I also think 99.9% may be a bit high of an estimate :)

Separately, we also need to have an easy way to set the queue on the shared instance of PINRemoteImage

danielgalasko:
So if I can outline the following requirements for the callback queue:

- It should *not* default to the main queue
- It should be configured on initialisation of the manager

Anything else before I hop back on?

garrettmoon:
I think there are two changes that are absolutely necessary:

1. It should not be mutable, and should be configured at initialization
2. There should be a straightforward way to configure the shared instance of the manager. I'm not sure the best way to achieve this… Essentially, in a very large app, you want to be able to ensure that `[PINRemoteImageManager sharedImageManager]` uses a queue you specify. Hopefully without having to set it in `+load` or `+initialize`

I wouldn't say that it's required to default to non-main (I see the benefits you suggest, I just question the drawbacks), though yet one other issue: if we change the default to main, the developer cost will be much higher to change back to non-main if we ever choose to do so – developers will have assumed that callbacks happen on main throughout their app and will either have to change the queue back to main (hopefully they read the release notes) or audit all uses of PINRemoteImage.

garrettmoon:
I'm going to close this pull request since I haven't heard anything back for a while.

Issue #158, opened 2016-03-08
Cocoapods 1.0 Beta 5 + WebP + use_frameworks!


smyrgl:
When testing out Cocoapods 1.0 (Beta 5) the issue with frameworks that 0ef39a8b7f599112c8944a7c8b921be28ac8dc9e works around seems to have cropped up again.

<img width="1260" alt="screen shot 2016-03-08 at 9 33 35 am" src="https://cloud.githubusercontent.com/assets/1709178/13610087/e2f816fe-e510-11e5-8025-e6903341bb4f.png">

I'm not sure what change they made in 1.0 that broke this workaround but worth bringing this up now so alternate ways to explore linking this in can be looked at.

bdorfman:
Also just wanted to mention that the commit mentioned (0ef39a8b7f599112c8944a7c8b921be28ac8dc9e) appears to totally break Carthage support for using WebP with PINRemoteImage since it now relies on a preprocessor macro being set that is only set through Cocoapods. If you revisist framework linking, you should fix this as well.

garrettmoon:
@smyrgl can you post your pod file or a minimal version so I can test? The release candidate for 1.0 just came out so I think now is the time to address this :)

garrettmoon:
@smyrgl Can you reproduce this on the 1.0.0 release of CocoaPods? I'm not able to. Please ping this thread if you're able to and I'll re-open the issue.

yxztj:
@garrettmoon I'm using Cocoapods 1.0.1 and also facing this issue. It's easy to repro.
Repro steps:
1. Create an empty project
2. Create a podfile as follows:
`
target 'testPIN' do
  use_frameworks!
  pod 'PINRemoteImage'
  pod 'PINRemoteImage/WebP'
end`
3. pod install
4. open the newly created workspace, and build, an error pops up saying:
￼
![1 pic](https://cloud.githubusercontent.com/assets/3334458/17094108/bdd3ec18-527f-11e6-9031-375b487cae44.jpg)
￼

garrettmoon:
There appears to be an issue with the webp cocoa pod :( I don't think this is something that can actually be fixed on our end, but needs to be addressed in the webp repo. As far as I can tell, this will occur with *any* cocoapods project that has the webp repo as a dependency.

I would suggest opening an issue here: https://bugs.chromium.org/p/webp/issues/list

FYI, it still appears to work with 0.39.0

yxztj:
[...]

PR #226, opened 2016-08-04
Fixes webp integration for CocoaPods > 1.0. Thanks to SDWebImage for …

garrettmoon:
…the solution and @yxztj to pointing it out!

Addresses #158


yxztj:
Thanks @garrettmoon !

PR #229, opened 2016-08-09
Don't bother archiving / unarchiving data.

garrettmoon:
This change upgrades to PINCache 3.0.1-beta which supports
custom archiving and unarchiving.

This allows us to avoid using NSArchiver/Archiver which adds a
small amount of disk space overhead and between .5 and 10 ms (on an iPhone 4)
for each encode / decode.

To support this, we *do* have to add versioning for the cache and
destroy it for the old versions. I considered adding a process to migrate
the images, but because that would have to occur on startup and process
all existing images, it could cause issues for very large image cache
databases, so I decided against that.


garrettmoon:
Also added
- (void)removeCachedObjectForKeyFromMemoryCache:(NSString *)key;
to the image caching protocol

Example/PINRemoteImage Tests/PINRemoteImage_Tests.m
```diff
@@ -511,7 +512,16 @@ - (void)testload
 
 - (void)testInvalidObject
 {
-    [self.imageManager.cache cacheObjectOnDisk:@"invalid" forKey:[self.imageManager cacheKeyForURL:[self JPEGURL] processorKey:nil]];
+    NSString * const kPINRemoteImageDiskCacheName = @"PINRemoteImageManagerCache";
+    NSString *cacheURL = [NSSearchPathForDirectoriesInDomains(NSCachesDirectory, NSUserDomainMask, YES) objectAtIndex:0];
```
chrisdanford: `cachePath`?  The return value of this is a local file system path and not a URL.

Example/Pods/PINCache/PINCache/PINCache.h
```diff
@@ -94,18 +94,47 @@ typedef void (^PINCacheObjectContainmentBlock)(BOOL containsObject);
  @param name The name of the cache.
  @result A new cache with the specified name.
  */
-- (instancetype)initWithName:(NSString *)name;
+- (instancetype)initWithName:(nonnull NSString *)name;
```
chrisdanford: This is assumed to be non-null because of `NS_ASSUME_NONNULL_BEGIN`, right?  https://developer.apple.com/swift/blog/?id=25

Same for a few places below.

Example/PINRemoteImage Tests/PINRemoteImage_Tests.m
chrisdanford: Style nit: Consider `firstObject` instead of `objectAtIndex:0`.

Example/Pods/PINCache/PINCache/PINCache.h
garrettmoon: Yeah, this is actually in PINCache and I my guess is someone was just trying to be consistent when they put in the nullable serializer / deserializer. I didn't catch there was a NS_ASSUME above.

PR #233, opened 2016-08-19
Fixed completion handler crash

frowing:


garrettmoon:
Thanks for the PR @frowing!

Out of curiosity have you seen this crash in the wild? The header indicates that completionHandler should be non null (it has an NS_ASSUME_NONNULL_BEGIN at the top), wondering if it's an OS bug? If so, a comment would be great above the completionHandler checks, something along the lines of, "Even though this is documented to be non-nil, in the wild it sometimes is".

Also, I don't even understand how the system could continue if the completionHandler is nil. How would it know how to continue the request? Could there actually be a different issue here?

frowing:
@garrettmoon yep I've started seeing this on iOS `9.3.3` it might very well be an iOS issue. Here's one stack trace. 

```
Thread 15 Crashed:
0   libdispatch.dylib                    0x21b78db2 dispatch_async + 66
1   libsystem_blocks.dylib               0x21bbc95b _Block_copy_internal + 296
2   Flagship                             0x0030da77 -[PINURLSessionManager URLSession:task:didReceiveChallenge:completionHandler:] (PINURLSessionManager.m:79)
3   CFNetwork                            0x225d2187 __68-[NSURLSession delegate_task:didReceiveChallenge:completionHandler:]_block_invoke + 36
4   Foundation                           0x227fa84d __NSBLOCKOPERATION_IS_CALLING_OUT_TO_A_BLOCK__ + 6
5   Foundation                           0x2275b87f -[NSBlockOperation main] + 144
6   Foundation                           0x2274df8f -[__NSOperationInternal _start:] + 772
7   Foundation                           0x227fcb0d __NSOQSchedule_f + 190
8   libdispatch.dylib                    0x21b81e7f _dispatch_queue_drain + 1760
9   libdispatch.dylib                    0x21b7ae17 _dispatch_queue_invoke + 280
10  libdispatch.dylib                    0x21b8320d _dispatch_root_queue_drain + 398
11  libdispatch.dylib                    0x21b8307b _dispatch_worker_thread3 + 92
12  libsystem_pthread.dylib              0x21d16e0d _pthread_wqthread + 1022
13  libsystem_pthread.dylib              0x21d169fc start_wqthread + 6
```

Comment added :)

garrettmoon:
Thank you!

frowing:
@garrettmoon any idea when this fix will be released?

garrettmoon:
@frowing it will be in the next 3.0 beta for sure, but that probably won't be out for a bit. I'd recommend pointing to a fork or a commit hash if you need the fix immediately.

PR #234, opened 2016-08-21
Compatibility with network debugging framework 

huajiahen:
PINURLSessionManager currently can not work with network debugging framework such as FLEX or PonyDebugger, because these frameworks will hook NSURLSessionDataDelegate method `-URLSession:dataTask:didReceiveResponse:completionHandler:`. According to Apple's document, delegate who implement this method must call completionHandler or system will not proceed the request. 
Though it is those framework's duty to avoid hook this function when delegate didn't implement this, by adding this method will quick fix compatibility.

### DIFF
--- a/Pod/Classes/PINURLSessionManager.m
+++ b/Pod/Classes/PINURLSessionManager.m
@@ -72,6 +72,11 @@ NSString * const PINURLErrorDomain = @"PINURLErrorDomain";
 
 #pragma mark NSURLSessionDataDelegate
 
+- (void)URLSession:(NSURLSession *)session dataTask:(NSURLSessionDataTask *)dataTask didReceiveResponse:(NSURLResponse *)response completionHandler:(void (^)(NSURLSessionResponseDisposition))completionHandler
+{
+    completionHandler(NSURLSessionResponseAllow);
+}
+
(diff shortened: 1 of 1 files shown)


Pod/Classes/PINURLSessionManager.m
```diff
@@ -72,6 +72,11 @@ - (void)unlock
 
 #pragma mark NSURLSessionDataDelegate
 
+- (void)URLSession:(NSURLSession *)session dataTask:(NSURLSessionDataTask *)dataTask didReceiveResponse:(NSURLResponse *)response completionHandler:(void (^)(NSURLSessionResponseDisposition))completionHandler
```
garrettmoon: Can you add a comment explaining why this is implemented? That way someone won't remove this in the future?

huajiahen:
It's due to a bug in FLEX and PonyDebugger, only affect NSURLSession's delegates that didn't implement this method. Since most networking library such as AFNetworking/Alamofire/SDWebImage had implement this method, PINRemoteImage is one of few library which affected by this issue. Now they had fixed it in https://github.com/Flipboard/FLEX/pull/151, and I will close this pull request.

garrettmoon:
Oh cool, thanks for putting up the PR! Sorry I've been remiss in reviewing it quickly.

Issue #239, opened 2016-09-05
Feature request: support Base64 inlined images


k06a:
Some `NSURL`s can contain inlined images like this:

[...]

PR #240, opened 2016-09-05
Add support for in-url base64 encoded images

k06a:
Feature is described here: https://github.com/pinterest/PINRemoteImage/issues/239

### DIFF
--- a/Pod/Classes/PINRemoteImageManager.m
+++ b/Pod/Classes/PINRemoteImageManager.m
@@ -520,7 +520,8 @@ static dispatch_once_t sharedDispatchToken;
                 object = [UIImage imageWithData:data];
             }
             if (object) {
-                [self.cache.memoryCache setObject:object forKey:key];
+                // Uncomment when cache key will be hash of URL
+                //[self.cache.memoryCache setObject:object forKey:key];
                 if ([self earlyReturnWithOptions:options url:url object:object completion:completion]) {
                     return nil;
                 }
--- a/Pod/Classes/PINRemoteImageManager.m
+++ b/Pod/Classes/PINRemoteImageManager.m
@@ -513,7 +513,9 @@ static dispatch_once_t sharedDispatchToken;
         if (data) {
             id object;
             if ([url.resourceSpecifier hasPrefix:@"image/gif"]) {
+#if USE_FLANIMATED_IMAGE
                 object = [FLAnimatedImage animatedImageWithGIFData:data];
+#endif
             } else {
                 object = [UIImage imageWithData:data];
             }
--- a/Pod/Classes/PINRemoteImageManager.m
+++ b/Pod/Classes/PINRemoteImageManager.m
@@ -509,26 +509,16 @@ static dispatch_once_t sharedDispatchToken;
     }
     
     if ([url.absoluteString hasPrefix:@"data:"]) {
-        NSUInteger colonLocation = [url.absoluteString rangeOfString:@":"].location;
-        NSUInteger semicolonLocation = [url.absoluteString rangeOfString:@";"].location;
-        NSUInteger commaLocation = [url.absoluteString rangeOfString:@","].location;
-        if (colonLocation != NSNotFound && semicolonLocation != NSNotFound && commaLocation != NSNotFound && colonLocation < semicolonLocation && semicolonLocation < commaLocation) {
-            NSString *mediaType = [url.absoluteString substringWithRange:NSMakeRange(colonLocation + 1, semicolonLocation - colonLocation - 1)];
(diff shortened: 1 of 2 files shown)


Adlai-Holler:
@k06a Great feature! NSURL will actually do a lot of this work for us (rough impl):

```
if ([url.scheme isEqualToString:@"data"]) {
   NSData *data = [[NSData alloc] initWithContentsOfURL:url];
   // Check if nil and bail
   if ([url.resourceSpecifier hasPrefix:@"image/gif"]) { . . . } // kUTTypeGIF requires MobileCoreServices
}
```

k06a:
@Adlai-Holler wow, thanks!

k06a:
@Adlai-Holler really `url.scheme` is nil :(

Pod/Classes/PINRemoteImageManager.m
```diff
@@ -508,6 +508,21 @@ - (NSUUID *)downloadImageWithURL:(NSURL *)url
         }
     }
     
+    if ([url.absoluteString hasPrefix:@"data:"]) {
+        NSData *data = [NSData dataWithContentsOfURL:url];
+        if (data) {
+            id object;
+            if ([url.resourceSpecifier isEqualToString:@"image/gif"]) {
```
Adlai-Holler: Should be "hasPrefix:"

Pod/Classes/PINRemoteImageManager.m
```diff
@@ -508,6 +508,21 @@ - (NSUUID *)downloadImageWithURL:(NSURL *)url
         }
     }
     
+    if ([url.absoluteString hasPrefix:@"data:"]) {
```
Adlai-Holler: Should be `url.scheme isEqualToString`

Pod/Classes/PINRemoteImageManager.m
k06a: And what is possible suffix?

Pod/Classes/PINRemoteImageManager.m
k06a: `url.scheme` is actually `nil`

Adlai-Holler:
Thanks for the diff! Could you add a unit test or two for this functionality? And did this not work beforehand? Loading data URLs via the normal URL loading system should work fine.

Pod/Classes/PINRemoteImageManager.m
Adlai-Holler: Hmm I'm seeing scheme="data" using the example URL you posted.

k06a:
Yep, I'll add some tests for this cases. Before this changes – such urls really not worked.

Pod/Classes/PINRemoteImageManager.m
Adlai-Holler: The resourceSpecifier will contain the data as well. It's everything after "data:"

k06a:
@Adlai-Holler do you think we need to add image to memory cache to avoid mutiple image decoding?

Pod/Classes/PINRemoteImageManager.m
k06a: My fail

Pod/Classes/PINRemoteImageManager.m
k06a: My fail

appleguy: Hi Anton - thanks for implementing this feature!

For caching, I think it's an important question.  Using these URLs as cache keys will increase the size of the cache, as the compressed image will itself be contained in the key.  This might be OK for apps that choose to use this feature, but something to consider.

k06a: @appleguy  maybe **not** to cache until next major release, which will use hash of url for cache key?

k06a:
@Adlai-Holler added test with image from this article https://en.wikipedia.org/wiki/Data_URI_scheme

Adlai-Holler:
@k06a Could you remove the caching for now, and maybe we'll add it back later? Otherwise I think this is ready to go!

k06a:
@Adlai-Holler just commented it. Can I hope to see this changes in `2.1.5` or `2.2.0`?

Adlai-Holler:
@k06a Thanks for adding this! I'm not sure which version the change will be available in yet, sorry. We'll have to talk with @garrettmoon and see what he has in mind.

Adlai-Holler:
Oh! @k06a would you be willing to reopen this pull request to target the `master` branch instead of the release branch?

k06a:
@Adlai-Hoover do you wanna me just reopen PR to another branch or rebase it from `master`?

Adlai-Holler:
@k06a Reopen the PR with `master` set as the base branch please.

k06a:
https://github.com/pinterest/PINRemoteImage/pull/245

PR #241, opened 2016-09-06
Ensure Opaque Images Have No Alpha Channel

Adlai-Holler:
~~It seems that even when downloading plain old JPEGs, we always return images with an alpha channel which is detrimental to performance.~~ 

~~I'm going to try and fix it!~~

Pairs nicely with https://github.com/facebook/AsyncDisplayKit/pull/2197

**EDIT** I'm dumb. But unit tests are good!


Adlai-Holler:
Nope I'm a moron. `kCGImageAlphaNoneSkipLast` is the same as none. We good.

Adlai-Holler:
So in the end this diff just adds some unit tests. Never hurts!

PR #242, opened 2016-09-06
Use kCGBlendModeCopy when Decoding Images

Adlai-Holler:
This makes image decoding faster.


Adlai-Holler:
@appleguy @garrettmoon Could someone give this a look?

appleguy:
Looks good to me - even if the image has transparency, copying in the pixels will still preserve that.  Since the context itself does not have content to blend with, this should be safe / functionally identical.

PR #255, opened 2016-10-13
Update image cache naming before public

garrettmoon:


garrettmoon:
@skeeet What do you think about these new names?

skeeet:
Looks more consistent with Foundation naming conventions. +1

PR #261, opened 2016-10-25
[PINAnimatedImageManager] Refine file cleanup in +initialize.

appleguy:
Question: is +initialize sufficient to perform this work?  Since the class may not be initialized,
even on the next launch of the app, do we need to more proactively clean up (without impacting app start?).

One possibility would be to listen for app launch and suspend notifications rather than +load.

cc @billkunz @garrettmoon

### DIFF
--- a/Pod/Classes/PINAnimatedImageManager.m
+++ b/Pod/Classes/PINAnimatedImageManager.m
@@ -39,15 +39,14 @@ BOOL PINStatusCoverImageCompleted(PINAnimatedImageStatus status) {
 
 @end
 
-static dispatch_once_t startupCleanupOnce;
-
 @implementation PINAnimatedImageManager
 
 + (void)initialize
 {
   if (self == [PINAnimatedImageManager class]) {
-    dispatch_async(dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_DEFAULT, 0), ^{
-      dispatch_once(&startupCleanupOnce, ^{
+    static dispatch_once_t startupCleanupOnce;
+    dispatch_once(&startupCleanupOnce, ^{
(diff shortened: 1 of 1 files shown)


Pod/Classes/PINAnimatedImageManager.m
```diff
@@ -39,15 +39,14 @@ + (instancetype)sharedManager;
...
 @implementation PINAnimatedImageManager
 
 + (void)initialize
 {
   if (self == [PINAnimatedImageManager class]) {
-    dispatch_async(dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_DEFAULT, 0), ^{
-      dispatch_once(&startupCleanupOnce, ^{
+    static dispatch_once_t startupCleanupOnce;
+    dispatch_once(&startupCleanupOnce, ^{
+      dispatch_async(dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_DEFAULT, 0), ^{
```
garrettmoon: This doesn't seem safe to run at an arbitrary point in the future?

garrettmoon:
With regard to your question, yeah, I'm not sure if initialize is enough, good point. Probably need something akin to what PINCache does for it's disk initialize that runs on app launch.

Pod/Classes/PINAnimatedImageManager.m
garrettmoon: Wait! This was already running on an arbitrary queue :( Sorry, I misunderstood the previous code!

garrettmoon: That explains why GIF pins don't restore correctly…

appleguy:
Feel free to either merge, or to close / abandon this if it's not the right approach.  I'm not too sure, just based this on a refinement of the existing code without changing the behavior.

Issue #123, opened 2016-01-29
How does the cache expire? Is it possible to check if a cached image has changed?


dev-mush:
I find PINRemoteImage a great and stable library, and I really thank you guys for such a great work!

There's just one thing I can't seem to figure out, and I promise I'll dig down into the source code to figure it out, but meanwhile, I'd like to ask if this functionality is actually supported.
( _I did some googling about it and crunched your past issues, and even if I found some similar questions, I didn't manage to find something that was answering my question_).

Sometimes, even if URL of the image remains the same, the Image on the backend changes. I'm wondering if PINRemoteImage checks http cache headers to understand if the image has expired. As far  as I've understood it seems that, as long as the url key is cached by PINCache, the cached image is retrieved from the disk and the manager doesn't try any download, is this correct?

If so, is there a way to workaround it? Like by setting cache expiration or forcing a header check and redownloading?

garrettmoon:
@dev-mush you're right, this is not currently supported by PINRemoteImage. There's currently a pull request attempting to implement TTL on PINCache, that could help. Looking at headers would certainly be a good option as well, the downloading of images is done in PINDataTaskOperation if you want to take a look on what it would take to add support.

garrettmoon:
@dev-mush TTL is now supported on PINCache, does this resolve your issue?

garrettmoon:
Closing this out since I haven't heard back.

PR #272, opened 2016-11-11
Add option to bypass cache

zachwaugh:
Related to https://github.com/pinterest/PINRemoteImage/issues/123, we need a way to expire the cache for images that have the same url, but the content changes. Since PINRemoteImage doesn't have a way of handling this automatically, we need some way of handling this ourselves.

For that, I added a new `PINRemoteImageManagerDownloadOptionsIgnoreCache` option. When that is set, it will bypass the memory and disk cache checks and force the image to be downloaded. This way we can load from cache, but periodically force an update based on external factors. I'm not very familiar with the internals, so let me know if there is a better way to achieve this, but wanted to get a conversation going. Thanks!

### DIFF
--- a/Pod/Classes/PINRemoteImageManager.m
+++ b/Pod/Classes/PINRemoteImageManager.m
@@ -564,12 +564,14 @@ static dispatch_once_t sharedDispatchToken;
         UUID = [NSUUID UUID];
     }
 
-    //Check to see if the image is in memory cache and we're on the main thread.
-    //If so, special case this to avoid flashing the UI
-    id object = [self.cache objectFromMemoryForKey:key];
-    if (object) {
-        if ([self earlyReturnWithOptions:options url:url key:key object:object completion:completion]) {
-            return nil;
+    if ((options & PINRemoteImageManagerDownloadOptionsIgnoreCache) == 0) {
+        //Check to see if the image is in memory cache and we're on the main thread.
+        //If so, special case this to avoid flashing the UI
+        id object = [self.cache objectFromMemoryForKey:key];
+        if (object) {
+            if ([self earlyReturnWithOptions:options url:url key:key object:object completion:completion]) {
+                return nil;
+            }
         }
     }
     
@@ -1520,6 +1522,11 @@ static dispatch_once_t sharedDispatchToken;
 
 - (void)objectForKey:(NSString *)key options:(PINRemoteImageManagerDownloadOptions)options completion:(void (^)(BOOL found, BOOL valid, PINImage *image, id alternativeRepresentation))completion
 {
+    if (options & PINRemoteImageManagerDownloadOptionsIgnoreCache) {
+        completion(NO, NO, nil, nil);
+        return;
+    }
+
     void (^materialize)(id object) = ^(id object) {
         PINImage *image = nil;
         id alternativeRepresentation = nil;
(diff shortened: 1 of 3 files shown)


garrettmoon:
@zachwaugh do you mind rebasing your PR? We just changed the CI for PINRemoteImage and you'll need the latest master for it to work.

zachwaugh:
@garrettmoon sure thing, just rebased and pushed

Pod/Classes/PINRemoteImageManager.m
```diff
@@ -1520,6 +1522,11 @@ - (NSString *)cacheKeyForURL:(NSURL *)url processorKey:(NSString *)processorKey
 
 - (void)objectForKey:(NSString *)key options:(PINRemoteImageManagerDownloadOptions)options completion:(void (^)(BOOL found, BOOL valid, PINImage *image, id alternativeRepresentation))completion
 {
+    if (options & PINRemoteImageManagerDownloadOptionsIgnoreCache) {
```
garrettmoon: total nit, but do you mind matching the style above, wrapping this in parens and checking to see that it != 0 ?

Pod/Classes/PINRemoteImageManager.m
```diff
@@ -1520,6 +1522,11 @@ - (NSString *)cacheKeyForURL:(NSURL *)url processorKey:(NSString *)processorKey
...
+        completion(NO, NO, nil, nil);
```
garrettmoon: I think the second param should be YES? Otherwise PINRemoteImage will report errors about invalid items in the cache.

Pod/Classes/PINRemoteImageManager.m
```diff
@@ -1520,6 +1522,11 @@ - (NSString *)cacheKeyForURL:(NSURL *)url processorKey:(NSString *)processorKey
 
 - (void)objectForKey:(NSString *)key options:(PINRemoteImageManagerDownloadOptions)options completion:(void (^)(BOOL found, BOOL valid, PINImage *image, id alternativeRepresentation))completion
 {
+    if ((options & PINRemoteImageManagerDownloadOptionsIgnoreCache) != 0) {
+        completion(NO, NO, nil, nil);
```
garrettmoon: I think this got missed, the second param here should be YES , otherwise PINRemoteImage will report an error to the client.

zachwaugh: 👍  I'll fix that as well, I matched that values here for when the object isn't in the cache - https://github.com/pinterest/PINRemoteImage/blob/master/Pod/Classes/PINRemoteImageManager.m#L1545.

It looks like we'd have the same problem here, right? https://github.com/pinterest/PINRemoteImage/blob/master/Pod/Classes/PINRemoteImageManager.m#L1100.

garrettmoon: You're absolutely correct! Mind fixing it there too? Good catch!

zachwaugh: Sure! So just return, `completion(NO, YES, nil, nil)` in both cases then?

garrettmoon: Exactly

garrettmoon:
And even a unit test! Thanks @zachwaugh !

PR #273, opened 2016-11-14
Add PINRemoteImageCaching to public framework headers

zachwaugh:
Fixes PINRemoteImageManager's cache property not being accessible from Swift:

image


garrettmoon:
Thanks @zachwaugh !

PR #276, opened 2016-11-15
Update PINRemoteImageManager.h

rahul-malik:
Documentation is incorrect. The NSURLSessionConfiguration is actually the ephemeral session if one is not specified.


garrettmoon:
Thanks @rahul-malik !

Issue #260, opened 2016-10-24
Images loaded from Disk Cache occasionally display artifacts / corruption.


baby03201:
I'm currently using AsyncDisplayKit + PINRemoteImage libraries in my app.
The PINRemoteImage version is 3.0.0-beta.3 (due to some constraint using in AsyncDisplayKit)

This library is awesome, however, sometimes the cached images look not fully downloaded as the image below:
![broken](https://cloud.githubusercontent.com/assets/923188/19639675/fed3c2be-9a0b-11e6-802b-17ce8b2923bc.jpeg)

The function I'm using is 

```
UIImageView *imageView = [[UIImageView alloc] init];
[imageView setPin_updateWithProgress:YES];
[imageView pin_setImageFromURL:[NSURL URLWithString:@"$(my_image_link)"]];
```

garrettmoon:
@baby03201 I have a guess as to the cause of that:
Are you running in the simulator. I've seen similar issues when running in the simulator, but *never* on device.

appleguy:
It is possible that there is an encoding error or other issue with the image file.  Could you send the URL to the image of that size?

baby03201:
@garrettmoon Unfortunately, I couldn't able to regenerate that error in my device now. This screenshot was taken by my friend with his iPhone 6s device. I guess it might happens while the network connection is bad.

@appleguy Thanks for your patience and response!
The images URL link was shown as below:
[First image](http://resource.golface.com.tw/ADPlatform/image/ALBA%E9%9B%9C%E8%AA%8C/2016%E5%B9%B49%E6%9C%88/alba-mag-cover201609.jpg)
[Second image](http://resource.golface.com.tw/ADPlatform/image/ALBA%E9%9B%9C%E8%AA%8C/2016%E5%B9%B48%E6%9C%88/alba-mag-cover201608.jpg)

SteveGeyer:
I came here looking specifically for information this bug. We just upgraded AsyncDisplayKit from 1.9.81 to 1.9.91 and similarly got PINRemoteImage upgraded to 3.0.0-beta.3. We we started seeing bad images on live devices (not the simulator) and we had not previously been seeing them in the three months we have been using this product. We started focussing on PINRemoteImage when we realize that the problem occurred with images fully managed and rendered by AsyncDisplayKit as well as by code where we called PINRemoteImage directly to load other images.

[...]

PR #279, opened 2016-12-06
Need to clear out existing progress data on retry

garrettmoon:
Finally found it! When retrying a download, we need to clear out the old fetched data, otherwise we'll append the new data on top of the old. This addresses #260 260


garrettmoon:
Thank you so much to everyone who helped track this down, especially @jaichaudhry !

PR #282, opened 2016-12-07
Move examples to their own directory

### LINKED ISSUE #281: Examples should be moved to their own directory

hannahmbanana:
Resolves #281.

This diff is messy to look at, so here's the steps I took:
1. moved example projects to a new Examples folder
2. removed podfile.lock, Pods/ directory and workspace file from each example
3. updated build script to point to the examples folder & run pod install

Build results below - does this look normal? 
screen shot 2016-12-07 at 1 37 59 pm


garrettmoon:
Awesome, thank you @hannahmbanana !

PR #285, opened 2016-12-08
Refactor Not to Use allKeys

### LINKED ISSUE #283: [Performance] Refactor use of -allKeys for a 1-2% global CPU optimization.
In this trace, -allKeys and the retain / release costs account for >1% of global app CPU (in this case, a brief scroll in the Profile screen of Pinterest).

Considering that this trace includes substantial image decoding / drawing costs, other UIKit / UICollectionView costs, and also heavy cache reading and writing — this is a surprising % total for this call.

As a simple solution, it would probably be worth maintaining the keys as a separate, duplicated array alongside the NSDictionary to avoid this.  Alternatively, NSHashMap or a similar class might allow for accessing the keys as an array directly without a copy.

A fix for this would further extend PINRemoteImage's performance lead on many competing image downloaders :)

screen shot 2016-12-07 at 11 38 38 pm

Adlai-Holler:
Resolves #283 cc @appleguy

### DIFF
--- a/Pod/Classes/PINRemoteImageManager.m
+++ b/Pod/Classes/PINRemoteImageManager.m
@@ -1036,13 +1036,10 @@ static dispatch_once_t sharedDispatchToken;
         typeof(self) strongSelf = weakSelf;
         [strongSelf lock];
             PINRemoteImageTask *taskToEvaluate = nil;
-            for (NSString *key in [strongSelf.tasks allKeys]) {
-                PINRemoteImageTask *task = [strongSelf.tasks objectForKey:key];
-                for (NSUUID *blockUUID in [task.callbackBlocks allKeys]) {
-                    if ([blockUUID isEqual:UUID]) {
-                        taskToEvaluate = task;
-                        break;
-                    }
+            for (PINRemoteImageTask *task in [strongSelf.tasks objectEnumerator]) {
+                if (task.callbackBlocks[UUID] != nil) {
+                    taskToEvaluate = task;
+                    break;
                 }
             }
         
@@ -1062,16 +1059,13 @@ static dispatch_once_t sharedDispatchToken;
     [_concurrentOperationQueue pin_addOperationWithQueuePriority:PINRemoteImageManagerPriorityHigh block:^{
         typeof(self) strongSelf = weakSelf;
         [strongSelf lock];
-        for (NSString *key in [strongSelf.tasks allKeys]) {
-            PINRemoteImageTask *task = [strongSelf.tasks objectForKey:key];
-            for (NSUUID *blockUUID in [task.callbackBlocks allKeys]) {
-                if ([blockUUID isEqual:UUID]) {
-                    if ([task isKindOfClass:[PINRemoteImageDownloadTask class]]) {
-                        PINRemoteImageCallbacks *callbacks = task.callbackBlocks[blockUUID];
-                        callbacks.progressImageBlock = progressImageCallback;
-                    }
-                    break;
+        for (PINRemoteImageTask *task in [strongSelf.tasks objectEnumerator]) {
+            PINRemoteImageCallbacks *callbacks = task.callbackBlocks[UUID];
+            if (callbacks != nil) {
+                if ([task isKindOfClass:[PINRemoteImageDownloadTask class]]) {
+                    callbacks.progressImageBlock = progressImageCallback;
                 }
+                break;
             }
         }
         [strongSelf unlock];
--- a/Pod/Classes/PINRemoteImageManager.m
+++ b/Pod/Classes/PINRemoteImageManager.m
@@ -1543,32 +1543,11 @@ static dispatch_once_t sharedDispatchToken;
 /// Attempt to find the task with the callbacks for the given uuid
 - (nullable PINRemoteImageTask *)_locked_taskForUUID:(NSUUID *)uuid key:(NSString * _Nullable * _Nullable)outKey
 {
-    /**
-     * Use enumerateKeysAndObjects concurrently to lookup the task.
-     * This method uses dispatch_apply under the hood and so is safe
-     * from thread explosion.
-     *
-     * Block-based enumeration should be avoided in general, but unfortunately
-     * using NSDictionary.objectEnumerator + fast enumeration results in a
-     * slowish path involving 2N calls to -nextObject, N calls to objectForKey:,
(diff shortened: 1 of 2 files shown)


appleguy:
This is great, thanks Adlai!  I also approve :).  Laughing a bit about the simple NSDictionary objectForKey: replacing a full iteration of allKeys and checking each one :).

Adlai-Holler:
@jparise @appleguy I did something controversial with the latest commit. As I say in the comments, obviously block-based enumeration is satan the devil but in this case it's the best way to go. Thoughts?

Pod/Classes/PINRemoteImageManager.m
```diff
@@ -999,29 +1000,21 @@ - (void)cancelTaskWithUUID:(NSUUID *)UUID
...
-            if ([taskToEvaluate cancelWithUUID:UUID manager:strongSelf]) {
-                [strongSelf.tasks removeObjectForKey:taskKey];
-            }
-        [strongSelf unlock];
+         typeof(self) strongSelf = weakSelf;
+         [strongSelf lock];
+             NSString *taskKey = nil;
+             PINRemoteImageTask *taskToEvaluate = [strongSelf _locked_taskForUUID:UUID key:&taskKey];
+
+         if (taskToEvaluate == nil) {
```
jparise: Indentation looks off here. Should this be indented within the -lock/-unlock? Or is GitHub just making it hard to see?

Pod/Classes/PINRemoteImageManager.m
```diff
@@ -1564,6 +1540,48 @@ - (void)objectForKey:(NSString *)key options:(PINRemoteImageManagerDownloadOptio
...
+    [self.tasks enumerateKeysAndObjectsWithOptions:NSEnumerationConcurrent usingBlock:^(NSString * _Nonnull key, __kindof PINRemoteImageTask * _Nonnull task, BOOL * _Nonnull stop) {
+        // If this isn't our task, or if we've already found it (somehow!), just return.
+        if (task.callbackBlocks[uuid] == nil || atomic_flag_test_and_set(&finished)) {
+            return;
+        }
+
+        // Found it! Save our results and end enumeration
+        result = task;
+        if (outKey != NULL) {
+            *outKey = key;
```
jparise: Why do this inside of the block? It would be simpler to do this just before we return `result` below. That would be one less variable for the block to capture, too.

Pod/Classes/PINRemoteImageManager.m
```diff
@@ -1035,18 +1028,8 @@ - (void)setPriority:(PINRemoteImageManagerPriority)priority ofTaskWithUUID:(NSUU
     [_concurrentOperationQueue pin_addOperationWithQueuePriority:PINRemoteImageManagerPriorityHigh block:^{
         typeof(self) strongSelf = weakSelf;
         [strongSelf lock];
-            PINRemoteImageTask *taskToEvaluate = nil;
-            for (NSString *key in [strongSelf.tasks allKeys]) {
-                PINRemoteImageTask *task = [strongSelf.tasks objectForKey:key];
-                for (NSUUID *blockUUID in [task.callbackBlocks allKeys]) {
```
garrettmoon: Oh man, definitely didn't need allKeys here…

Pod/Classes/PINRemoteImageManager.m
```diff
@@ -1564,6 +1540,48 @@ - (void)objectForKey:(NSString *)key options:(PINRemoteImageManagerDownloadOptio
...
+     * A flag to protect our result.
+     * In practice there will only be one task that passes this test and so
+     * the chances of us concurrently writing to result are virtually 0 but
+     * the cost here is also virtually 0 so let's just be safe.
+     */
+    __block atomic_flag finished = ATOMIC_FLAG_INIT;
+
+    [self.tasks enumerateKeysAndObjectsWithOptions:NSEnumerationConcurrent usingBlock:^(NSString * _Nonnull key, __kindof PINRemoteImageTask * _Nonnull task, BOOL * _Nonnull stop) {
+        // If this isn't our task, or if we've already found it (somehow!), just return.
+        if (task.callbackBlocks[uuid] == nil || atomic_flag_test_and_set(&finished)) {
```
garrettmoon: This is a bit tricks-y? Would this be clearer if we had two ifs, one following the other?

Pod/Classes/PINRemoteImageManager.m
```diff
@@ -1564,6 +1540,48 @@ - (void)objectForKey:(NSString *)key options:(PINRemoteImageManagerDownloadOptio
...
+
+    /**
+     * A flag to protect our result.
+     * In practice there will only be one task that passes this test and so
+     * the chances of us concurrently writing to result are virtually 0 but
+     * the cost here is also virtually 0 so let's just be safe.
+     */
+    __block atomic_flag finished = ATOMIC_FLAG_INIT;
+
+    [self.tasks enumerateKeysAndObjectsWithOptions:NSEnumerationConcurrent usingBlock:^(NSString * _Nonnull key, __kindof PINRemoteImageTask * _Nonnull task, BOOL * _Nonnull stop) {
```
garrettmoon: Are we sure that the concurrent enumeration allows a performance improvement here? The work of each block is so small?

Adlai-Holler: @appleguy thanks for the response!

Have you looked at my comment added in this diff and the article it links? -objectEnumerator is quite slow, it's even slower than using -objectForKey: for each key. All those extra message sends and the extra object allocation will take a major toll.

Would you be willing to profile this implementation?

Adlai-Holler: Eh on second thought it won't matter much either way. String hashing is crazy fast and the imp-cache will be active so probably not worth either of our times to profile. Removing the extra enumeration and allKeys will be the lion's share

Pod/Classes/PINRemoteImageManager.m
Adlai-Holler: We won't have access to the key below, and so the block will have to capture some variable either way. This way results in one less variable – would have to create a local variable otherwise – so let's stick with it.

Pod/Classes/PINRemoteImageManager.m
Adlai-Holler: Indentation is correct here – according to xcode it was off by 1 char before.

PR #287, opened 2016-12-11
Make Some Low-Hanging Optimizations

Adlai-Holler:
Nothing too special. These changes are safe and reduce the number of calls going into the runtime.


appleguy:
@Adlai-Holler very nice, thanks for looking at this!  PINRemoteImage is still one of the most leveraged areas we can optimize for Pinterest (based on current profiles), and surely will benefit other clients too.

appleguy:
@garrettmoon could you approve the build and land this once it passes?  Changes look good to me.

garrettmoon:
@Adlai-Holler @appleguy the builds will be automatically approved if you set your membership of the Pinterest org to public here: https://github.com/orgs/pinterest/people (find yourself and click next to the lock).

PINRemoteImage/PINRemoteImage.xcodeproj/project.pbxproj
```diff
@@ -425,7 +425,7 @@
 				GCC_WARN_UNINITIALIZED_AUTOS = YES_AGGRESSIVE;
 				GCC_WARN_UNUSED_FUNCTION = YES;
 				GCC_WARN_UNUSED_VARIABLE = YES;
-				IPHONEOS_DEPLOYMENT_TARGET = 8.0;
+				IPHONEOS_DEPLOYMENT_TARGET = 7.0;
```
garrettmoon: Good catch for Carthage users!

Pod/Classes/PINRemoteImageMacros.h
```diff
@@ -35,10 +35,14 @@
 #define PINImage     UIImage
 #define PINImageView UIImageView
 #define PINButton    UIButton
+#define PINNSOperationSupportsQOS (kCFCoreFoundationVersionNumber >= kCFCoreFoundationVersionNumber_iOS_8_0)
+#define PINNSURLSessionTaskSupportsPriority (kCFCoreFoundationVersionNumber >= kCFCoreFoundationVersionNumber_iOS_8_0)
```
garrettmoon: Wait, this is compile time, no? That means if will always evaluate to true as long as it's compiled with a version of the SDK that's greater than or equal to 8. However, when it's run on iOS 7, it will still be true and cause crashes?

appleguy:
Interesting note on the runtime checks -- I think you're right, the code is probably invalid as changed.  It should be replaced with a static BOOL / dispatch_once arrangement.

Adlai-Holler:
This isn't compile-time. See the AS_AT_LEAST macros. @garrettmoon @appleguy

Adlai-Holler:
@garrettmoon Thanks for the pointer, I've publicized my membership in the org.

garrettmoon:
@Adlai-Holler Look at that! This is way better then performance wise.

To get CI to pass we need to either come up with a solution so that the project has a deployment target of iOS 8 when built with Carthage or to leave it as is. Really, we should probably have a separate Carthage project than use the example :/ Let's do that. Can you revert the change to the deployment target and I'll create this issue: (will edit)

Adlai-Holler:
@garrettmoon Awesome! Reverted the deployment target change. Here we go!

PR #293, opened 2016-12-24
Add ability to configure header after initialize PINRemoteImageManager

sainttail:
This will come in handy when we want to access secure image which usually require us to provide `Authorization` header.

### DIFF
--- a/Pod/Classes/PINRemoteImageManager.h
+++ b/Pod/Classes/PINRemoteImageManager.h
@@ -184,10 +184,10 @@ typedef void(^PINRemoteImageManagerProgressDownload)(int64_t completedBytes, int
 - (nonnull id<PINRemoteImageCaching>)defaultImageCache;
 
 /**
- * Set the custom header directly to every requests. Headers set from this method will override any header from NSURLSessionConfiguration.
+ * Sets a custom header to be included in every request. Headers set from this method will override any header from NSURLSessionConfiguration.
  *
- * @param value A value for the header
- * @param header  A string field for header
+ * @param value A value for the header. Pass in nil to remove a previously set value.
+ * @param header A string field for header.
  */
 - (void)setValue:(nullable NSString *)value forHTTPHeaderField:(nullable NSString *)header;
 
--- a/Pod/Classes/PINRemoteImageManager.h
+++ b/Pod/Classes/PINRemoteImageManager.h
@@ -184,6 +184,14 @@ typedef void(^PINRemoteImageManagerProgressDownload)(int64_t completedBytes, int
 - (nonnull id<PINRemoteImageCaching>)defaultImageCache;
 
 /**
+ * Set the custom header directly to every requests. Headers set from this method will override any header from NSURLSessionConfiguration.
+ *
+ * @param value A value for the header
+ * @param header  A string field for header
+ */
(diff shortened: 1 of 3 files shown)


Pod/Classes/PINRemoteImageManager.h
```diff
@@ -184,6 +184,14 @@ typedef void(^PINRemoteImageManagerProgressDownload)(int64_t completedBytes, int
 - (nonnull id<PINRemoteImageCaching>)defaultImageCache;
 
 /**
+ * Set the custom header directly to every requests. Headers set from this method will override any header from NSURLSessionConfiguration.
```
garrettmoon: Could the first sentence say "Sets a custom header to be included in every request"

Pod/Classes/PINRemoteImageManager.h
```diff
@@ -184,6 +184,14 @@ typedef void(^PINRemoteImageManagerProgressDownload)(int64_t completedBytes, int
...
+ *
+ * @param value A value for the header
```
garrettmoon: Can you add: Pass in nil to remove a previously set value.

garrettmoon:
@sainttail Sorry this has taken so long to get in, do you mind rebasing against master?

sainttail:
@garrettmoon Done!

Pod/Classes/PINRemoteImageManager.m
```diff
@@ -281,6 +283,21 @@ - (void)unlock
     [_lock unlock];
 }
 
+- (void)setValue:(nullable NSString *)value forHTTPHeaderField:(nullable NSString *)header {
+    __weak typeof(self) weakSelf = self;
+    dispatch_async(dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_DEFAULT, 0), ^{
+        typeof(self) strongSelf = weakSelf;
+        [strongSelf lock];
+            if (value) {
+                strongSelf.httpHeaderFields[header] = value;
```
garrettmoon: Do you think we should change this to `strongSelf.httpHeaderFields[[header copy]] = [value copy];`? That way if it turns out someone has passed in a mutable string and changes it later we won't get in trouble?

Pod/Classes/PINRemoteImageManager.m
```diff
@@ -281,6 +283,21 @@ - (void)unlock
...
+            }
+            else {
+                [strongSelf.httpHeaderFields removeObjectForKey:header];
```
garrettmoon: Do we need the else here? I was under the impression NSMutableDictionary's behavior was to remove the object if you set it to nil using the object subscripting stuff?

garrettmoon:
Sorry for all the nits, ready to land once these are addressed :)

Pod/Classes/PINRemoteImageManager.m
sainttail: You're right. Thank for pointing this out.

Pod/Classes/PINRemoteImageManager.m
sainttail: Yes That's line is not need looking for doc, I have just known that setting to nil remove the object instead!

sainttail:
@garrettmoon Fixed, thank for suggestion!

garrettmoon:
Thanks @sainttail for adding this feature!

PR #295, opened 2016-12-31
[WIP] Support Buck build

nguyenhuy:
First attempt to support Buck. Only library targets are included for now (no test targets yet).

How to use:
```bash
# Run this once on the root directory to resolve dependencies (i.e PINCache and FLAnimatedImage) 
# and populate their BUCK files (more on this later)
pod install
buck build lib
# Same as above because 'lib' is an alias for //:PINRemoteImage
buck build //:PINRemoteImage
buck build //:PINRemoteImage-Core
```

Dependency resolution:
  - Currently I build dependencies from source files downloaded by CocoaPods.
  - There is a `post_install` hook added to the Podfile in root dir. The hook copies all BUCK templates needed to build the pods. Our main BUCK file (in the root dir) has hardcoded paths that point to these dependent BUCK targets. That means `pod install` needs to be run at least once.
  - Buck can already build static and shared prebuilt frameworks (`prebuilt_apple_framework`). So we can rely on Carthage for dependency resolution if needed to. I used it to build FLAnimatedImage for a while, but decided abandoned this approach because `prebuilt_apple_framework` is still under development and poorly documented.
 - BUCK templates have just enough information to build source files downloaded by CocoaPods. That means it either is an exact copy of the BUCK file from original repo (in case of PINCache), or is a stripped down version of the original BUCK file (in case of PINRemoteImage itself when added to AsyncDisplayKit), or only lives in our repo (in case of FLAnimatedImage, at least for now).

Let me know if you have any questions or suggestions!


maicki:
@garrettmoon @nguyenhuy What do you think about that PR? Do we still want to add buck support to PINRemoteImage? If so maybe we should rebase soon and get it as we are currently fighting with CI anyway ... Thoughts?

garrettmoon:
Yeah, we should try to bring this home or close the PR.

nguyenhuy:
I'd prefer to switch to Basel since we may be able to use it internally sooner or later. In addition, since Basel supports git dependencies, I'll be able to remove the CocoaPods hooks. Let's close this PR.

Issue #284, opened 2016-12-08
How do you manually insert an image into the cache?


zachwaugh:
I have an image that I fetch from an API that's easier to grab directly than through PINRemoteImage since it requires authentication. Once I have that image, how I can insert it directly into the shared cache? I see there is the `PINRemoteImageCaching` protocol that has write methods, but it isn't clear exactly how to use it. What do I insert into the cache, the raw `NSData`? Do I insert into the memory or disk cache?

This looks like it would work, but wanted to double check:

```swift
let data = ... // NSData from HTTP request
let imageManager = PINRemoteImageManager.shared()
let cacheKey = imageManager.cacheKey(for: url, processorKey: nil)
imageManager.cache.setObjectInMemory(data, forKey: cacheKey, withCost: 0)
```

Thanks!

garrettmoon:
@zachwaugh hmm, this isn't actually well supported now, but we absolutely could support it. Essentially we'd want to add a method to PINRemoteImage which is something like:

`insertImageDataIntoCache:(nonull NSData*)data withURL:(nonull NSURL *)url processingKey:(nullable NSString *)processingKey`

This would build and insert a PINRemoteImageMemoryContainer and insert into the memoryCache and  would store the data directly into the diskCache.

As a workaround, you could store the data directly to the disk cache with `cacheKeyForURL:processorKey:` and `setObjectOnDisk:forKey:` The disadvantage of this is it will need to be loaded into the memoryCache from disk on the next access.

Hope this helps! Going to close this, but please feel free to reply back if it's not clear.

zachwaugh:
@garrettmoon thanks! I'll take a look at adding that functionality and opening a PR

garrettmoon:
Awesome, thanks @zachwaugh !

PR #299, opened 2017-01-06
Add an image manually to the cache.

danielrhodes:
I needed to add an image manually to the cache and saw this issue referenced here: https://github.com/pinterest/PINRemoteImage/issues/284

It seems to work, but is there anything I am missing?

### DIFF
--- a/Pod/Classes/PINRemoteImageManager.h
+++ b/Pod/Classes/PINRemoteImageManager.h
@@ -424,6 +424,22 @@ typedef void(^PINRemoteImageManagerProgressDownload)(int64_t completedBytes, int
                              progressImage:(nullable PINRemoteImageManagerImageCompletion)progressImage
                                 completion:(nullable PINRemoteImageManagerImageCompletion)completion;
 
+
+/**
+ Adds an image manually into the memory and disk cache.
+ 
+ @param data NSData with the raw image data.
+ @param url NSURL where the image resides.
+ @processorKey NSString key to uniquely identify processor and process. Will be used for caching processed images.
+ @additionalCost NSUInteger the additional cost (for cache eviction purposes) to generate the processed image
+ 
+ @return A BOOL indicating if the image was successfully added to the cache.
+ */
+- (BOOL) insertImageDataIntoCache:(nonnull NSData*)data
+                          withURL:(nonnull NSURL *)url
+                     processorKey:(nullable NSString *)processorKey
+                   additionalCost:(NSUInteger)additionalCost;
+
 /**
  Returns the cacheKey for a given URL and processorKey. Exposed to be overridden if necessary or to be used with imageFromCacheWithCacheKey
  @see imageFromCacheWithCacheKey:completion:
--- a/Pod/Classes/PINRemoteImageManager.m
+++ b/Pod/Classes/PINRemoteImageManager.m
@@ -751,6 +751,25 @@ static dispatch_once_t sharedDispatchToken;
     [self unlock];
 }
 
+-(BOOL) insertImageDataIntoCache:(nonnull NSData*)data
+                         withURL:(nonnull NSURL *)url
+                    processorKey:(nullable NSString *)processorKey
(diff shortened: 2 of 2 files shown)


Pod/Classes/PINRemoteImageManager.m
```diff
@@ -751,6 +751,25 @@ - (void)downloadImageWithURL:(NSURL *)url
...
+-(BOOL) insertImageDataIntoCache:(nonnull NSData*)data withURL:(nonnull NSURL *)url processorKey:(nullable NSString *)processorKey {
+  
+  if (url != nil) {
+    NSString *key = [self cacheKeyForURL:url processorKey:processorKey];
+    
+    PINRemoteImageManagerDownloadOptions options = PINRemoteImageManagerDownloadOptionsSkipDecode & PINRemoteImageManagerDownloadOptionsSkipEarlyCheck;
+    PINRemoteImageMemoryContainer *container = [[PINRemoteImageMemoryContainer alloc] init];
+    container.data = data;
+    
+    return [self materializeAndCacheObject: container cacheInDisk: data additionalCost:0 key:key options:options outImage: nil outAltRep: nil];
```
garrettmoon: I think additionalCost is used when dealing with processed images. If this method is going to take a URL and a processorKey, it should also probably take a processingCost parameter and pass that in to the additional cost?

Pod/Classes/PINRemoteImageManager.m
```diff
@@ -751,6 +751,25 @@ - (void)downloadImageWithURL:(NSURL *)url
...
+  }
+  
+  return NO;
+}
+
+- (BOOL) insertImageDataIntoCache:(NSData *)data withURL:(NSURL *)url {
```
garrettmoon: I'm not sure it's worth exposing this convenience method as it's not likely to be called that often?

Pod/Classes/PINRemoteImageManager.m
danielrhodes: Just to clarify on what cost is here: this is a number which can be set arbitrarily high which would increase the likelihood that an object gets evicted from the cache?

Pod/Classes/PINRemoteImageManager.h
```diff
@@ -424,6 +424,17 @@ typedef void(^PINRemoteImageManagerProgressDownload)(int64_t completedBytes, int
...
+
+/**
+ Adds an image manually into the memory and disk cache.
+ 
+ @param data NSData with the raw image data.
+ @param url NSURL where the image resides.
+ @processorKey NSString key to uniquely identify processor and process. Will be used for caching processed images.
+ @additionalCost NSUInteger The cache cost of the object.
+ */
+- (BOOL) insertImageDataIntoCache:(nonnull NSData*)data withURL:(nonnull NSURL *)url processorKey:(nullable NSString *)processorKey additionalCost: (NSUInteger) additionalCost;
```
garrettmoon: One more tiny nit: do you think this would be better exposed as processingCost? I'm not entirely sure, I think that might make more sense. Also super nit: there's a space in front of additionalCost.

danielrhodes: @garrettmoon Based on how it is ultimately used here: https://github.com/pinterest/PINRemoteImage/blob/master/Pod/Classes/PINRemoteImageManager.m#L1432, `additionalCost` or `additionalCacheCost` would be most accurate. I think `additionalCacheCost` is best, as I was initially confused on what it does.

No problem on removing that space 😃

garrettmoon: Ok, lets go with additionalCost, can you update the comment to say `@additionalCost NSUInteger the additional cost (for cache evicting purposes) to generate the processed image`

danielrhodes: Done.

danielrhodes:
I'll squash these commits later. ^

garrettmoon:
No need to squash, GitHub makes that easy :) Thanks for the contribution!

PR #302, opened 2017-01-20
Fix race condition

garrettmoon:


Pod/Classes/PINAnimatedImageManager.m
```diff
@@ -113,7 +129,7 @@ - (void)animatedPathForImageData:(NSData *)animatedImageData infoCompletion:(PIN
   __block BOOL startProcessing = NO;
   __block PINSharedAnimatedImage *sharedAnimatedImage = nil;
   {
-    [_lock lockWithBlock:^{
+    [_lock lockWhenCondition:PINAnimatedImageManagerConditionReady];
```
master-nevi: `animatedPathForImageData:infoCompletion:completion:` can be called on the main thread. You should probably dispatch to a background queue before locking.

Pod/Classes/PINAnimatedImageManager.m
```diff
@@ -41,15 +46,22 @@ + (instancetype)sharedManager;
 
 @implementation PINAnimatedImageManager
 
-+ (void)initialize
++ (void)load
```
master-nevi: Usually better practice for a 3rd party library to use `initialize` just because it's only called as soon as the first message to the class is sent not when the class is loaded at launch.

garrettmoon: Yeah, so the intention was to clear out any images even if no new GIFs were loaded. I tried to make this as lightweight as possible by only registering for a notification.

Wonder if there's an alternative to load to force the code to run at some point after launch…

garrettmoon: I guess I could dispatch_async this, that's probably even more lightweight.

garrettmoon: For the record I did profile this and it didn't show up in the trace I did.

garrettmoon: Traced again with dispatch_async version and that also doesn't show up. Will go with that.

master-nevi: Wait, this is to clear images from the temporary directory? Correct me if I'm wrong but I thought the temp directory is always cleared by the system for every run of the app.

garrettmoon: I've never been able to find documentation on when it's cleared but it's definitely not on every run of the app (I just confirmed to be sure).

Pod/Classes/PINAnimatedImageManager.m
```diff
@@ -105,6 +120,7 @@ - (instancetype)init
 
 + (void)cleanupFiles
 {
+  NSLog(@"cleaning files…");
```
master-nevi: PINRemoteImage doesn't have a logging system?

Pod/Classes/PINAnimatedImageManager.m
```diff
@@ -41,14 +46,23 @@ + (instancetype)sharedManager;
...
-+ (void)initialize
++ (void)load
 {
   if (self == [PINAnimatedImageManager class]) {
-    static dispatch_once_t startupCleanupOnce;
-    dispatch_once(&startupCleanupOnce, ^{
-      dispatch_async(dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_DEFAULT, 0), ^{
-        [self cleanupFiles];
-      });
+    dispatch_async(dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_DEFAULT, 0), ^{
```
master-nevi: I guess my last comment is that depending on the device load, you may not be guaranteed that the dispatch queue block will run before the app `DidFinishLaunching`, so now you're not guaranteed that sharedManager will run when you think it will. I'd almost say it's better without the dispatch here.

master-nevi: You should probably also clean up your notification observation at init.

Pod/Classes/PINAnimatedImageManager.m
```diff
@@ -41,14 +46,23 @@ + (instancetype)sharedManager;
...
+#if PIN_TARGET_IOS
+      NSString * const notificationName = UIApplicationDidFinishLaunchingNotification;
+#elif PIN_TARGET_MAC
+      NSString * const notificationName = NSApplicationDidFinishLaunchingNotification;
+#endif
+      
+      [[NSNotificationCenter defaultCenter] addObserverForName:notificationName
```
master-nevi: You should probably also clean up your notification observation at init.

appleguy:
Just a quick note, I do think it is important that we do something like +load.  Imagine an app that uses a GIF somewhere deep in its UI, and then crashes (hopefully not due to this library — but for any reason, crashes).  When the app restarts, even if the user goes back there, it’s crucial to do the cleanup.  Agreed though it should be delayed maybe even 30 seconds after app launch with a dispatch_async.  I wonder if a background task is appropriate.

		— Scott

garrettmoon:
Thanks @master-nevi and @appleguy ! I think I'll just remove the on startup and dispatch after 10 seconds. Seems like a happy medium :) I don't think the background task is necessary, this is just removing a directory.

master-nevi:
👍

PR #306, opened 2017-02-02
Added build setting to support app extensions

shanev:
If PINRemoteImage is used inside an app extension, you get the linker warning: `linking against a dylib which is not safe for use in application extensions`. This adds the `APPLICATION_EXTENSION_API_ONLY` build setting to silence the warning.


shanev:
PINCache has been updated in `master` to allow app extensions, but isn't in a release yet. Once it's released, PINRemoteImage should automatically pick it up because the dependency is set to `>= 3.0.1`.

Issue #152, opened 2016-02-25
Image scale is always 1.0


rounak:
The images created by PINRemoteImage always has a scale of 1.0, which causes problems with intrinsic sizes of image views. SDWebImage create images with the correct scale factor depending on the device.

garrettmoon:
Thanks for the report @rounak, definitely something we should look into addressing!

rounak:
@garrettmoon This is the fix we added https://github.com/TinderApp/PINRemoteImage/commit/bea7c9d6636e392b6b2b5a0809bf13629b0d0fc9 We didn't use `UIScreen scale` because we weren't sure if that can be called from non main threads. Let me know if you think this is a good fix. If so, I can open up a PR

garrettmoon:
@rounak Thanks for bringing this up and figuring out a solution!

I think this should probably be done at a different layer, or specified at a different layer. All the images that are downloaded from the web don't have 'scale' per say as far as I know. So this would set the scale arbitrarily to the screen size. I'm not sure what the effects would be for people relying on the scale being 1.0 currently.

Since scale is generally used by UIImageView, it seems like this sort of abstraction would live better there. Would it be better to have the UIImage modified in pin_updateUIWithRemoteImageManagerResult in PINImageView+PINRemoteImage?

We could add a bool which indicates whether images should be 'scaled' and if so, create a new UIImage
```
NSAssert(resultImage.CGImage, @"UIImage must have CGImage backing");
[UIImage imageWithCGImage:[resultImage CGImage] scale:[self contentScaleFactor] orientation:[resultImage orientation]]
```

Alternatively, the scale could be something that is set on the PINRemoteImageManager, but I think it makes more sense to be part of the view's.

What do you think?

rounak:
@garrettmoon Agreed that the current layer where this is done seems a little off. However, if we just put it in the image view category, places where images are prefetched, or where we're directly using the image manager would be excluded. So I think PINRemoteImageManager seems like a better candidate to add that.

One thing to note is that UIImage size is measured in points and not pixels. So as per Apple's docs, to get the actual size in pixels of an image, you should always multiple size by scale.
![screen shot 2016-05-02 at 1 02 52 pm](https://cloud.githubusercontent.com/assets/240258/14966241/392bc82e-1066-11e6-9108-15ceeaa268d7.png)

matrixx:
Hi! Just bumped into the same issue. What's the status of the upcoming fix?

[...]

PR #308, opened 2017-02-09
Support for high resolution images

maicki:
The problem is that currently PINRemoteImage in `pin_decodedImageWithData:skipDecodeIfPossible` is decoding image data always with a scale of 1.0. This is most of the time the right way as images are served in 1x. The problem appears if retina images are served from the server like @2x or @3x images. The decoding of this image data will produce not the right result.

This PR adds support for automatic scaling of images there are served if image URLs contain a @2x or @3x.

cc #152

### DIFF
--- a/Examples/Example/PINRemoteImage Tests/PINRemoteImage_Tests.m
+++ b/Examples/Example/PINRemoteImage Tests/PINRemoteImage_Tests.m
@@ -136,6 +136,11 @@ static inline BOOL PINImageAlphaInfoIsOpaque(CGImageAlphaInfo info) {
     return [NSURL URLWithString:@"https://placekitten.com/g/200/301?longarg=helloMomHowAreYouDoing.IamFineJustMovedToLiveWithANiceChapWeTravelTogetherInHisBlueBoxThroughSpaceAndTimeMaybeYouveMetHimAlready.YesterdayWeMetACultureOfPeopleWithTentaclesWhoSingWithAVeryCelestialVoice.SoGood.SeeYouSoon.MaybeYesterday.WhoKnows.XOXO"];
 }
 
+- (NSURL *)scaledImageURL
+{
+    return [NSURL URLWithString:@"http://www.grenadabluewatersailing.com/wp-content/uploads/2014/09/Coconut-wikipedia.com_-300x225@2x.jpg"];
+}
+
 #pragma mark - <PINURLSessionManagerDelegate>
 
 - (void)didReceiveData:(NSData *)data forTask:(NSURLSessionTask *)task
@@ -418,6 +423,22 @@ static inline BOOL PINImageAlphaInfoIsOpaque(CGImageAlphaInfo info) {
     XCTAssert(outImageEncoded && [outImageEncoded isKindOfClass:[UIImage class]], @"Failed downloading image or image is not a UIImage.");
     XCTAssert(outImageDecoded && [outImageDecoded isKindOfClass:[UIImage class]], @"Failed downloading image or image is not a UIImage.");
     XCTAssert(encodedDrawTime / decodedDrawTime > 2, @"Drawing decoded image should be much faster");
+}
+
+- (void)testScalingWithURLPostfix
+{
+    
+    XCTestExpectation *expectation = [self expectationWithDescription:@"Error on scaling"];
+    [self.imageManager downloadImageWithURL:[self scaledImageURL]
+                                    options:PINRemoteImageManagerDownloadOptionsNone
+                                 completion:^(PINRemoteImageManagerResult *result)
+     {
+         XCTAssert(result.image != nil);
+         XCTAssert(result.image.scale == 2.0);
+         
+         [expectation fulfill];
+    }];
+    [self waitForExpectationsWithTimeout:[self timeoutTimeInterval] handler:nil];
 }
 
 - (void)drawImage:(UIImage *)image
--- a/Examples/Example/PINRemoteImage Tests/PINRemoteImage_Tests.m
+++ b/Examples/Example/PINRemoteImage Tests/PINRemoteImage_Tests.m
@@ -640,7 +640,7 @@ static inline BOOL PINImageAlphaInfoIsOpaque(CGImageAlphaInfo info) {
 {
     XCTestExpectation *expectation = [self expectationWithDescription:@"Image from cache"];
 
-    [self.imageManager imageFromCacheWithURL:[self JPEGURL] options:PINRemoteImageManagerDownloadOptionsNone completion:^(PINRemoteImageManagerResult * _Nonnull result) {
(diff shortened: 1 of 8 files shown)


Pod/Classes/Categories/PINImage+ScaledImage.m
```diff
@@ -0,0 +1,53 @@
...
+//  UIImage+ScaledImage.m
+//  Pods
+//
+//  Created by Michael Schneider on 2/9/17.
+//
+//
+
+#import "PINImage+ScaledImage.h"
+
+inline PINImage *PINScaledImageForKey(NSString * __nullable key, PINImage * __nullable image) {
```
jparise: Why `inline`? You'll end up duplicating this medium-sized function body twice below, increasing the size of the executable. Seems this can be `static` without a real performance cost.

maicki: Thanks! We had the function as export as well as I used it initially only one time. Removed the export for now and made it static we don't use it outside.

Examples/Example/PINRemoteImage Tests/PINRemoteImage_Tests.m
```diff
@@ -136,6 +136,11 @@ - (NSURL *)veryLongURL
     return [NSURL URLWithString:@"https://placekitten.com/g/200/301?longarg=helloMomHowAreYouDoing.IamFineJustMovedToLiveWithANiceChapWeTravelTogetherInHisBlueBoxThroughSpaceAndTimeMaybeYouveMetHimAlready.YesterdayWeMetACultureOfPeopleWithTentaclesWhoSingWithAVeryCelestialVoice.SoGood.SeeYouSoon.MaybeYesterday.WhoKnows.XOXO"];
 }
 
+- (NSURL *)scaledImageURL
+{
+    return [NSURL URLWithString:@"http://www.grenadabluewatersailing.com/wp-content/uploads/2014/09/Coconut-wikipedia.com_-300x225@2x.jpg"];
```
garrettmoon: Let's DDOS my server instead: http://excitedpixel.com/img/logo@2x.png

README.md
```diff
@@ -176,6 +176,22 @@ PINRemoteImageManager.shared().setAuthenticationChallenge { (task, challenge, co
 }
 ```
 
+### Support for high resolution images
+Currently there are two ways PINRemoteImage is supporting high resolution images:
+1. If the URL will contains a `@2x` or a `@3x` postfix it will automatically handled from PINRemoteImage and the resulting image will be returned with the right scale.
```
garrettmoon: nit: If the URL contains an `@2x` or an `@3x` postfix it will be automatically handled by PINRemoteImage and the resulting image will be returned at the right scale.

README.md
```diff
@@ -176,6 +176,22 @@ PINRemoteImageManager.shared().setAuthenticationChallenge { (task, challenge, co
...
+2. If it's not possible to provide an URL with the `@2x` or `@3x` postfix, it also can be handled within the completion handler:
```
garrettmoon: nit: If it's not possible to provide an URL with an `@2x` or `@3x` postfix, you can also handle it with a completion handler:

Pod/Classes/PINRemoteImageManager.m
```diff
@@ -1452,6 +1453,7 @@ - (BOOL)materializeAndCacheObject:(id)object
         }];
         if (image == nil) {
             image = [PINImage pin_decodedImageWithData:container.data skipDecodeIfPossible:skipDecode];
+            image = [PINImage pin_scaledImageForImage:image withKey:key];
```
garrettmoon: This won't work for processed images I don't think as the key will be an combination of the URL and a processing key. Either we need a method to extract the URL from the key or we need to pass the URL down to this method :(

See `cacheKeyForURL:processorKey:`

All though the burden added by supporting processing images is huge, sometimes I wonder if we should kill it…

Pod/Classes/Categories/PINImage+ScaledImage.m
```diff
@@ -0,0 +1,53 @@
...
+//
+
+#import "PINImage+ScaledImage.h"
+
+static inline PINImage *PINScaledImageForKey(NSString * __nullable key, PINImage * __nullable image) {
+    if (image == nil) {
+        return nil;
+    }
+    
+#if PIN_TARGET_IOS
```
garrettmoon: We should probably assert if image.CGImage is nil here, it isn't guaranteed to be non-nil (though it always should be if created by PINRemoteImage)

Pod/Classes/Categories/PINImage+ScaledImage.m
```diff
@@ -0,0 +1,53 @@
...
+    CGFloat scale = 1.0;
+    if (key.length >= 8) {
+        NSRange range = [key rangeOfString:@"@2x."];
```
garrettmoon: Should we search backwards for the . and then check for 2x in front?

maicki:
After looking into this a bit more, we may should go a different route:

We should *not* use @2x in the filename, since the @ character in a URL is reserved as the delimiter separating the username and password from the host in authentication URL schemes.

Instead we should follow a common file-naming convention for high-resolution images and prepend `_2x` or `_3x` before the file extension; for example, `myImage_2x.jpg`.

garrettmoon:
Can we check for both? <--- this guy has URLs on his server with @2x in them :)

Pod/Classes/PINRemoteImageManager.h
```diff
@@ -490,6 +500,16 @@ typedef void(^PINRemoteImageManagerProgressDownload)(int64_t completedBytes, int
 - (nonnull PINRemoteImageManagerResult *)synchronousImageFromCacheWithCacheKey:(nonnull NSString *)cacheKey options:(PINRemoteImageManagerDownloadOptions)options;
```
garrettmoon: Should this be deprecated?

Pod/Classes/PINRemoteImageManager.h
```diff
@@ -490,6 +500,16 @@ typedef void(^PINRemoteImageManagerProgressDownload)(int64_t completedBytes, int
...
 
 /**
+ Directly get an image from the underlying memory cache synchronously.
+ 
+ @param url NSURL to look up image in the cache.
+ @param options options will be used to determine if the cached image should be decompressed or FLAnimatedImages should be returned.
+ 
+ @return A PINRemoteImageManagerResult
+ */
+- (nonnull PINRemoteImageManagerResult *)synchronousImageFromCacheWithURL:(NSURL *)url options:(PINRemoteImageManagerDownloadOptions)options;
```
garrettmoon: This should have processorKey too?

Pod/Classes/PINRemoteImageManager.h
maicki: Done

Pod/Classes/Categories/PINImage+ScaledImage.m
maicki: Added

PR #314, opened 2017-02-13
Move to makefile

garrettmoon:
This PR cleans up the repo, making the Carthage and tests target the same. It also switches to Carthage as the build system. Of note, based on my understanding of Carthage, it now builds the PINRemoteImage framework without linking to PINCache or PINOperation, because the expectation is that linking will be done in the app that includes PINRemoteImage.


Cartfile
```diff
@@ -1 +1,2 @@
-github "pinterest/PINCache" "3.0.1-beta.3"
\ No newline at end of file
+git "https://chromium.googlesource.com/webm/libwebp" "v0.5.1"
+github "pinterest/PINCache" "fixCarthage"
```
garrettmoon: I'll update this once https://github.com/pinterest/PINCache/pull/171 lands.

PINRemoteImage/PINRemoteImageTests/PINRemoteImageTests.m
```diff
@@ -2,8 +2,8 @@
 //  PINRemoteImageTests.m
 //  PINRemoteImageTests
 //
-//  Created by Michael Schneider on 1/3/16.
-//  Copyright © 2016 mischneider. All rights reserved.
+//  Created by Garrett Moon on 2/13/17.
```
garrettmoon: I don't know why this got changed one way or the other…

PR #316, opened 2017-02-17
Fix infinite loop in synchronousImageFromCache method

zachwaugh:
`[PINRemoteImageManager synchronousImageFromCacheWithURL:(NSURL *)url processorKey:(nullable NSString *)processorKey options:(PINRemoteImageManagerDownloadOptions)options]` currently calls itself, creating an infinite loop. I'm assuming it's meant to call the more specific method below.


garrettmoon:
Ah! Good catch, thank you @zachwaugh !

PR #318, opened 2017-02-18
Expose pin cache

garrettmoon:


garrettmoon:
Yeah, the reason I want it exposed without an import is for people to be more likely to find it. I've had multiple emails from people asking how to limit cache and they didn't seem to realize they could just caste the result of .cache.

Issue #211, opened 2016-06-20
[Feature Proposal] URLRequest configuration


dtorres:
I would like to manipulate some properties of the request before is sent to the session, specifically `allowsCellularAccess` but other developers may find this useful to authenticate their requests through custom HTTP headers.

Currently it is only possible to modify these values at the session level, applying them to every request.

Implementation ideas:
- Add a parameter to PINRemoteImageManager methods to supply a block to configure the request
- Convert the options argument to a dictionary where users can supply values to be applied afterwards

And of course retrofit the view categories to support this.

garrettmoon:
I like the idea of a block which can be set which will be called to configure the request. Something like the  `authenticationChallenge` block that currently exists.

dtorres:
Having a block property in the manager takes away all the context a user might need. And considering that the only input the user has is the URL, which is useless since at this point it could've come from anywhere in the app (or the network for that matter)

garrettmoon:
We could also pass in the UUID? The concern I have with passing it in as part of the operation is that multiple areas of the app could be requesting the same image but with different request parameters. PINRemoteImage will only generate one request for multiple requests of a URL. How would it reconcile which set of request parameters to use?

dtorres:
makes sense, and let the user reconcile according to their needs

garrettmoon:
Hey, this got completed! Going to close this out.

PR #321, opened 2017-02-21
Add handler to allow customizing a network request

zachwaugh:
I had a similar need to https://github.com/pinterest/PINRemoteImage/issues/211, where I want to customize request headers on a per request basis, so I added a request configuration block on `PINRemoteImageManager` as suggested in that thread. My specific use-case is I want to use PINRemoteImageManager to fetch authenticated images for me, which requires setting an `Authorization` header with our token. However, I don't want to send this for images to other domains, so I need it to be per request. This seemed like the most straightforward approach, but let me know if you think it should be implemented another way.

### DIFF
--- a/Source/Classes/PINRemoteImageManager.m
+++ b/Source/Classes/PINRemoteImageManager.m
@@ -139,6 +139,7 @@ typedef void (^PINRemoteImageManagerDataCompletion)(NSData *data, NSError *error
 @property (nonatomic, assign) float lowQualityBPSThreshold;
 @property (nonatomic, assign) BOOL shouldUpgradeLowQualityImages;
 @property (nonatomic, copy) PINRemoteImageManagerAuthenticationChallenge authenticationChallengeHandler;
+@property (nonatomic, copy) PINRemoteImageManagerRequestConfigurationHandler requestConfigurationHandler;
 @property (nonatomic, strong) NSMutableDictionary <NSString *, NSString *> *httpHeaderFields;
 #if DEBUG
 @property (nonatomic, assign) float currentBPS;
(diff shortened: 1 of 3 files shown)


Source/Classes/PINRemoteImageManager.m
```diff
@@ -916,6 +927,11 @@ - (PINDataTaskOperation *)downloadDataWithURL:(NSURL *)url
     if (self.httpHeaderFields.count > 0) {
         request.allHTTPHeaderFields = [self.httpHeaderFields copy];
     }
+    
+    if (self.requestConfigurationHandler) {
```
garrettmoon: I think we need to lock to access this var?

Source/Classes/PINRemoteImageManager.m
```diff
@@ -295,6 +296,16 @@ - (void)setValue:(nullable NSString *)value forHTTPHeaderField:(nullable NSStrin
     });
 }
 
+- (void)setRequestConfiguration:(PINRemoteImageManagerRequestConfigurationHandler)configurationBlock {
+    __weak typeof(self) weakSelf = self;
+    dispatch_async(dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_DEFAULT, 0), ^{
+        typeof(self) strongSelf = weakSelf;
+        [strongSelf lock];
+        strongSelf.requestConfigurationHandler = configurationBlock;
```
garrettmoon: Mind indenting here? I've *attempted* to do this everywhere there's a lock :)

maicki:
Hey @zachwaugh you mind addressing the comments from @garrettmoon? Would be great if we could have this functionality within `PINRemoteImage` pretty soon. Thanks!

garrettmoon:
Hey @zachwaugh we're going to close this for now just to keep the PR queue clean. I'll reopen if you get a chance to address comments.

iOSUser110:
Hey if I want to pass Custom http header as XAPIKEY , then how to do that

zachwaugh:
Sorry for not following up on this PR earlier. It doesn't look like you can re-open a PR after rebasing and force pushing, so I just created a new branch/PR here - https://github.com/pinterest/PINRemoteImage/pull/355

PR #326, opened 2017-03-03
Move away from operation queue which was simply in place to limit the…

garrettmoon:
… number of downloads.


Source/Classes/PINRemoteImageDownloadQueue.h
```diff
@@ -0,0 +1,37 @@
...
+
+@class PINURLSessionManager;
+
+NS_ASSUME_NONNULL_BEGIN
+
+typedef void (^PINRemoteImageDownloadCompletion)(NSURLResponse *response, NSError *error);
+
+@interface PINRemoteImageDownloadQueue : NSObject
+
+@property (nonatomic, assign) NSUInteger maximumNumberOfOperations;
```
appleguy: does this refer to maxConcurrentOperations?  Maybe better as maxConcurrentDownloads, or connectionSlots, or something else less generic than "operations".

max number feels a bit like things will be dropped from the queue after hitting that number.

Source/Classes/PINRemoteImageDownloadQueue.h
```diff
@@ -0,0 +1,37 @@
...
+
+- (instancetype)init NS_UNAVAILABLE;
+- (PINRemoteImageDownloadQueue *)initWithMaximumNumberOfOperations:(NSUInteger)maximumNumberOfOperations NS_DESIGNATED_INITIALIZER;
```
appleguy: Consider a +queueWithMax:... (but only if this suits the framework's API design patterns).  It would make the creation line fit a bit more elegantly with a long name.

Source/Classes/PINRemoteImageDownloadQueue.h
```diff
@@ -0,0 +1,37 @@
...
+
+- (NSURLSessionDataTask *)addDownloadWithSessionManager:(PINURLSessionManager *)sessionManager
+                                                request:(NSURLRequest *)request
+                                               priority:(PINRemoteImageManagerPriority)priority
+                                      completionHandler:(PINRemoteImageDownloadCompletion)completionHandler;
+
+- (void)dequeueDownload:(NSURLSessionDataTask *)downloadTask;
+
+- (void)setTaskQueuePriority:(NSURLSessionDataTask *)downloadTask priority:(PINRemoteImageManagerPriority)priority;
```
appleguy: Consider setQueuePriority:fordownloadTask: ?  The current name feels a slightly confusing as Priority: in the first argument is repeated in the second.

To make the task be passed first, not sure how to do that naming actually!

Source/Classes/PINRemoteImageDownloadQueue.h
```diff
@@ -0,0 +1,37 @@
...
+
+- (instancetype)init NS_UNAVAILABLE;
+- (PINRemoteImageDownloadQueue *)initWithMaximumNumberOfOperations:(NSUInteger)maximumNumberOfOperations NS_DESIGNATED_INITIALIZER;
+
+- (NSURLSessionDataTask *)addDownloadWithSessionManager:(PINURLSessionManager *)sessionManager
+                                                request:(NSURLRequest *)request
+                                               priority:(PINRemoteImageManagerPriority)priority
+                                      completionHandler:(PINRemoteImageDownloadCompletion)completionHandler;
+
+- (void)dequeueDownload:(NSURLSessionDataTask *)downloadTask;
```
appleguy: Is this effectively cancelDownloadTask: ?  Dequeue could potentially imply processing the item through the queue, such as performing it synchronously.

Source/Classes/PINRemoteImageDownloadQueue.m
```diff
@@ -0,0 +1,145 @@
...
+        _defaultPriorityQueuedOperations = [[NSMutableArray alloc] init];
+        _lowPriorityQueuedOperations = [[NSMutableArray alloc] init];
+    }
+    return self;
+}
+
+- (NSUInteger)maximumNumberOfOperations
+{
+    [self lock];
+        NSUInteger maximumNumberOfOperations = _maximumNumberOfOperations;
```
appleguy: For these scalar types, there's been some really surprisingly good results using std::atomic in ASLayoutElementStyle.  It would be the first such usage in the framework, and so is your call.  The upside is that complexity is low (it would be a net removal of lines of code), and help the framework fulfill a hardcore dedication to minimized overhead for the app using it.

Some code examples & profiles here: https://github.com/facebook/AsyncDisplayKit/pull/2413 . The profiles don't capture how much savings it has in blocking though, only runtime benefits.  The real advantage is that it causes the thread scheduler to not have to stop executing the thread and then need to re-start it later:


image
![]

Tests/PINRemoteImageTests.m
```diff
@@ -983,4 +998,47 @@ - (void)testThatDecodedTransparentPNGImageHasAlpha
 	[self waitForExpectationsWithTimeout:[self timeoutTimeInterval] handler:nil];
 }
 
+- (void)testMaximumNumberOfDownloads
+{
+    __block NSInteger count = 0;
+    NSUInteger totalDownloads = [self bigURLs].count;
```
appleguy: Awesome test, pretty sophisticated / comprehensive!

Source/Classes/PINRemoteImageDownloadQueue.h
garrettmoon: It's a bit different than cancel since this class is intended to *only* manage positioning in the queue itself. I.e. dequeuing doesn't actually cancel an inflight network request, that's left up to the caller. Same with priority: it sets the priority in which a task will get executed but does not pass this through to the networking layer, also left up to the caller. I should probably add documentation :P

Source/Classes/PINRemoteImageDownloadQueue.m
garrettmoon: Seems like a good performance improvement, but I want to avoid premature optimization at this time. Happy to address if it comes up in traces once this is in use.

garrettmoon:
Thanks for the review @appleguy, will address feedback and get landed. Next is replacing NSOperation entirely (code is done, just wanted to land this first).

Source/Classes/PINRemoteImageDownloadQueue.h
garrettmoon: Ah, I see what you mean now, will change to removeDownloadTaskFromQueue:

PR #329, opened 2017-03-10
Fix for download queue

garrettmoon:


pinterest-ios:
Build failure:
https://gist.githubusercontent.com/pinterest-ios/8955410b712f258b6c2adefe3c678376/raw/bcde797b575a042419469857688524e215ed2b2f/results.txt

Source/Classes/PINRemoteImageDownloadQueue.m
```diff
@@ -99,42 +100,58 @@ - (void)scheduleDownloadsIfNeeded
...
+                    break;
+                    
+                case PINRemoteImageManagerPriorityDefault:
+                    queue = _defaultPriorityQueuedOperations;
+                    break;
+                    
+                case PINRemoteImageManagerPriorityHigh:
+                    queue = _highPriorityQueuedOperations;
+                    break;
+            }
```
maicki: What is happening on the default case? Should we assert in case an invalid priority is passed in.

garrettmoon: Good call, will add.

PR #331, opened 2017-03-14
Resume

garrettmoon:


pinterest-ios:
Build failure:
https://gist.githubusercontent.com/pinterest-ios/141d194c2fee07a942b6fdd10a518c3b/raw/3a2dd238ab9055fb9ccc281676bd6350406b32c8/results.txt

Source/Classes/PINRemoteImageDownloadTask.h
```diff
@@ -8,13 +8,17 @@
...
 #import "PINProgressiveImage.h"
+#import "PINResume.h"
 
 @interface PINRemoteImageDownloadTask : PINRemoteImageTask
 
 @property (nonatomic, strong, nullable) NSURLSessionDataTask *urlSessionTask;
 @property (nonatomic, assign) CFTimeInterval sessionTaskStartTime;
 @property (nonatomic, assign) CFTimeInterval sessionTaskEndTime;
 @property (nonatomic, assign) BOOL hasProgressBlocks;
+@property (nonatomic, strong, nullable) NSString *ifRange;
```
maicki: Let's use `copy` for NSString?

Source/Classes/PINResume.h
```diff
@@ -0,0 +1,24 @@
...
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface PINResume : NSObject
+
+- (id)init NS_UNAVAILABLE;
++ (PINResume *)resumeData:(NSData *)resumeData ifRange:(NSString *)ifRange totalBytes:(long long)totalBytes;
+
+@property (nonatomic, strong, readonly) NSData *resumeData;
+@property (nonatomic, strong, readonly) NSString *ifRange;
```
maicki: We should use `copy` here.

Source/Classes/PINRemoteImageManager.m
```diff
@@ -881,15 +889,26 @@ + (BOOL)retriableError:(NSError *)remoteImageError
...
-                                         timeoutInterval:self.timeout];
-    if (self.httpHeaderFields.count > 0) {
-        request.allHTTPHeaderFields = [self.httpHeaderFields copy];
+                                                           cachePolicy:NSURLRequestReloadIgnoringLocalCacheData
+                                                       timeoutInterval:self.timeout];
+    
+    NSMutableDictionary *headers = [self.httpHeaderFields mutableCopy];
+    
+    if (resume) {
+        headers[@"If-Range"] = resume.ifRange;
```
maicki: Should we check here if the `ifRange` is null? I think it's declared as `nullable`.

Source/Classes/PINProgressiveImage.m
```diff
@@ -116,9 +117,14 @@ - (CFTimeInterval)startTime
     return startTime;
 }
 
-- (void)updateProgressiveImageWithData:(NSData *)data expectedNumberOfBytes:(int64_t)expectedNumberOfBytes
+- (void)updateProgressiveImageWithData:(nonnull NSData *)data expectedNumberOfBytes:(int64_t)expectedNumberOfBytes isResume:(BOOL)isResume
 {
     [self.lock lock];
+        if (isResume) {
+            NSAssert(self.mutableData == nil, @"If we're resuming, data shouldn't be setup yet.");
+            self.startingBytes = data.length;
```
maicki: Should we set the property with the lock held in this case? Does this class not follow like the naming convention yet: `_locked_...`?

garrettmoon: PINRemoteImage only has a couple places where the _locked_ convention is followed currently :(

Source/Classes/PINRemoteImageManager.m
garrettmoon: It's not nullable, I have NS_ASSUME_NONNULL_BEGIN at the beginning.

garrettmoon: I will add asserts in the init though.

Source/Classes/PINProgressiveImage.m
garrettmoon: We should build some magic debug macro that adds in code on DEBUG which automatically checks that locks are held when a method prefixed with _locked_ is called!

Source/Classes/PINRemoteImageManager.m
```diff
@@ -881,15 +889,26 @@ + (BOOL)retriableError:(NSError *)remoteImageError
...
+        headers[@"Range"] = [NSString stringWithFormat:@"bytes=%tu-", resume.resumeData.length];
+    }
+    
+    if (headers.count > 0) {
+        //copy is probably not necessary here since we created headers locally…
```
nguyenhuy: Agreed! Let's avoid copying here.

Source/Classes/PINURLSessionManager.h
```diff
@@ -13,6 +13,7 @@ extern NSString * __nonnull const PINURLErrorDomain;
 @protocol PINURLSessionManagerDelegate <NSObject>
 
 @required
+- (void)didReceiveResponse:(nonnull NSURLResponse *)response forTask:(nonnull NSURLSessionTask *)task;
```
nguyenhuy: Because of [this](https://github.com/pinterest/PINRemoteImage/pull/331/files#diff-ad3073891c160428009d09e3f7d1efbdR176), does it make sense for this method to be optional?

PR #333, opened 2017-03-17
Update documentation for url, cacheKey and processorKey in PINRemoteImageManager

maicki:


Source/Classes/PINRemoteImageManager.h
```diff
@@ -491,7 +491,7 @@ typedef void(^PINRemoteImageManagerProgressDownload)(int64_t completedBytes, int
  @deprecated
  @see synchronousImageFromCacheWithURL:processorKey:options:
  
- @param cacheKey NSString key to look up image in the cache.
+ @param cacheKey NSString key that was used to download image
```
garrettmoon: This isn't quite correct, it should be `cacheKey obtained from @c cacheKeyForURL:processorKey`

PR #336, opened 2017-03-21
Add option to skip retry on download error.

chenium:


emablekos:
This would be great. We have a specific area of our app where we attempt to load an optional image, but if it takes longer than the timeout we set, we need to move on without the image.

garrettmoon:
Sorry for the slow action on this one!

PR #338, opened 2017-03-27
Rename module.modulemap to PINRemoteImage.modulemap

justin:
This is to work around a compilation issue with Xcode 8. Pretty sure its related to this radar: http://www.openradar.me/30194818


nguyenhuy:
Is this the same issue we're seeing in https://github.com/facebook/AsyncDisplayKit/issues/3232?

maicki:
@nguyenhuy Yeah I think so. We should do the same in ASDK.

PR #339, opened 2017-03-28
Fix warning for logging PINRemoteImageManagerPriority

maicki:

PR #340, opened 2017-03-31
Suppress warning in format string.

timonus:
This PR suppresses a warning due to a format string using `%d` instead of `%lu`.

screen shot 2017-03-31 at 10 09 50 am

### DIFF
--- a/Source/Classes/PINRemoteImageDownloadQueue.m
+++ b/Source/Classes/PINRemoteImageDownloadQueue.m
@@ -150,7 +150,7 @@
                     break;
                     
                 default:
-                    NSAssert(NO, @"invalid priority: %d", priority);
+                    NSAssert(NO, @"invalid priority: %lu", priority);
                     break;
             }
             [queue addObject:downloadTask];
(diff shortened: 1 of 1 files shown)


Source/Classes/PINRemoteImageDownloadQueue.m
```diff
@@ -150,7 +150,7 @@ - (void)setQueuePriority:(PINRemoteImageManagerPriority)priority forTask:(NSURLS
                     break;
                     
                 default:
-                    NSAssert(NO, @"invalid priority: %d", priority);
+                    NSAssert(NO, @"invalid priority: %lu", priority);
```
bkase: I think we need `%tu` here actually. `%lu` is only correct when building for 64bit targets. On 32bit targets `%lu` will still have the warning.

Source: https://twitter.com/gparker/status/377910611453046784?lang=en

garrettmoon:
@timonus mind addressing @bkase 's comments and we'll merge?

timonus:
@garrettmoon @bkase You got it!

timonus:
@bkase all updated now.

garrettmoon:
Ah, sorry, didn't see this one which got merged already: https://github.com/pinterest/PINRemoteImage/pull/339

PR #341, opened 2017-04-06
avoid EXC_BAD_ACCESS when delegateQueue is nil

ANNotunzdY:


garrettmoon:
@ANNotunzdY delegateQueue should never be nil at this point. Are you sure this is actually what's causing crashes? We've seen crashes around this code but they only occur on iOS 9 which leads me to believe they're bugs in the operating system. Does this change fix crashes for you?

ANNotunzdY:
@garrettmoon
Yes, It crash (delegateQueue is nil) occurs in maybe iOS 9 devices only.
I confirmed fix crashes by this code.

maicki:
@garrettmoon @ANNotunzdY I think we should do some more testing and profiling what is really going on here. I would like to understand the underlying problem before we are adding some "nil" checks. Good question is, as @garrettmoon already pointed out, is the `delegateQueue` nil at all in this case? @garrettmoon Can you point me to the crashes we see internally. Thanks!

maicki:
@garrettmoon and me looked into that a bit more today:
- We are still not to 100% sure what is going on, but it seems the crash is due to the `delegateQueue` becomes nil at some point.
- The crashes are unfrequenlty.
- It crashes in two `NSURLSessionDataDelegate` methods: `URLSession:dataTask: didReceiveResponse` and `URLSession:dataTask:didReceiveData:`
- This could happen if the two `NSURLSessionDataDelegate` callbacks mentioned above are called out of order as we remove the `delegateQueue` from a local dictionary after `URLSession:task: didCompleteWithError:` is called.
- In the [NSURLSession documentation](https://developer.apple.com/library/content/documentation/Cocoa/Conceptual/URLLoadingSystem/NSURLSessionConcepts/NSURLSessionConcepts.html) is an interesting part about this two `NSURLSessionDataDelegate` callbacks in regards to multipart download, that we actually don't do within Pinterest ... so it's questionable if this is could be the cause due some internal NSURLSession bug:

> If the response is multipart encoded, the session may call the delegate’s didReceiveResponse method again, followed by zero or more additional didReceiveData calls. If this happens, go to step 7 (handling the didReceiveResponse call).

- This *only* seems to happen in devices on iOS 9. 

In case we would merge this PR, it would introduce some side effects as the delegate callbacks are not called anymore and we would end up in a limbo state.

garrettmoon:
Yeah, I'd like to understand *why* the delegateQueue is nil. I have a feeling that merging this would result in corrupt images and / or crashes in different parts of the library.

pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

postmechanical:
Seeing these crashes as well quite frequently on iOS 9.3.4 and 9.3.5. On iOS 10.3.2 simulator I often see the NSURLSession delegate methods invoked on app launch for a data task that was running in a previous app session. I wonder whether that could be part of the problem here. On app launch, if the system is still invoking delegate methods for a previously in flight data task then `delegateQueue` for that previous data task will definitely be `nil`. Going to give this PR a try.

Source/Classes/PINURLSessionManager.m
```diff
@@ -149,6 +152,8 @@ - (void)URLSession:(NSURLSession *)session task:(NSURLSessionTask *)task didComp
                                     userInfo:@{NSLocalizedDescriptionKey : @"HTTP Error Response."}];
         }
     }
+    
+    if (delegateQueue == nil) return;
```
garrettmoon: Do you mind adding this assert above this line and all the other nil checks?:
`NSAssert(delegateQueue != nil, @"There seems to be an issue in iOS 9 where this can be nil. If you can reliably reproduce hitting this, *please* open an issue: https://github.com/pinterest/PINRemoteImage/issues");`

Source/Classes/PINURLSessionManager.m
```diff
@@ -78,6 +78,7 @@ - (void)URLSession:(NSURLSession *)session dataTask:(NSURLSessionDataTask *)task
         dispatch_queue_t delegateQueue = self.delegateQueues[@(task.taskIdentifier)];
     [self unlock];
     
+    if (delegateQueue == nil) return;
```
garrettmoon: Have you seen this one in the wild? We've only seen it on `didCompleteWithError`: and `didReceiveData:`

Would you mind removing this check?

Source/Classes/PINURLSessionManager.m
```diff
@@ -109,6 +110,7 @@ - (void)URLSession:(NSURLSession *)session task:(NSURLSessionTask *)task didRece
         dispatch_queue_t delegateQueue = self.delegateQueues[@(task.taskIdentifier)];
     [self unlock];
 
+    if (delegateQueue == nil) return;
```
garrettmoon: Same here, I don't think it's necessary here?

garrettmoon:
Hey @ANNotunzdY thanks so much for putting this up! I put up a PR that hopefully addresses this issue and adds some asserts so that maybe we can learn to understand it.

PR #342, opened 2017-04-08
Fix for maxNumberOfConcurrentDownloads

xdream86:
I find a issue about maxNumberOfConcurrentDownloads. maxNumberOfConcurrentDownloads default is 10. When I use Charles app check network request.
I find the tasks download simultaneously always 4, not 10.
before
I modified example project for test: every UICollectionViewCell limit to (30,30), change pic to more big size. 

Finally I find HTTPMaximumConnectionsPerHost property of NSURLSessionConfiguration default is 4. 
In order meet max number of concurrent download image task , we need keep maxNumberOfConcurrentDownloads less than or equals to HTTPMaximumConnectionsPerHost.
Because Change NSURLSessionConfiguration will not affect urlsession after has been initialized. And larger maxNumberOfConcurrentDownloads will lead download task timeout when at weak network quality. So I set default 
HTTPMaximumConnectionsPerHost is 20.
After I fixed this issue. 
now


Source/Classes/PINRemoteImageManager.m
```diff
@@ -36,10 +36,10 @@
...
 
 
-#define PINRemoteImageManagerDefaultTimeout     30.0
-#define PINRemoteImageMaxRetries                3
-#define PINRemoteImageRetryDelayBase            4
-
+#define PINRemoteImageManagerDefaultTimeout            30.0
+#define PINRemoteImageMaxRetries                       3
+#define PINRemoteImageRetryDelayBase                   4
+#define PINRemoteImageHTTPMaximumConnectionsPerHost    20
```
garrettmoon: Does it make sense to set this to NSIntegerMax since it will be controlled by maxNumberOfConcurrentDownloads anyway?

garrettmoon:
(sorry it took so long to review this)

pinterest-ios:
Build failure:
https://gist.githubusercontent.com/pinterest-ios/383efac3e7ee2da506f040513b8e3762/raw/f035ed67ea3aeeb0174df28ecf6f67823c3d0b60/results.txt

pinterest-ios:
Build failure:
https://gist.githubusercontent.com/pinterest-ios/9983a491ca6b91f1ebb94b4e551e3caa/raw/90c3ea3fe595f479c3a61d9b659cf1cf507fdbf0/results.txt

garrettmoon:
@xdream86 would you mind also rebasing, the 404 failure has been fixed in master.

Source/Classes/PINRemoteImageManager.m
xdream86: If setting PINRemoteImageHTTPMaximumConnectionsPerHost equals NSIntegerMax, NSURLSession will stop request anything.

pinterest-ios:
Build failure:
https://gist.githubusercontent.com/pinterest-ios/2ec1a76adb3d7b217a744527a04367c1/raw/0c0fbbb0b76d13dedb510253cce6cdae25cea4c6/results.txt

xdream86:
I have rebase code.

Source/Classes/PINRemoteImageManager.m
garrettmoon: Should we at least make it a larger number like 1000?

xdream86: Finally， I changed the HTTPMaximumConnectionsPerHost  value to equals UINT16_MAX.  

I read the answer from SO: http://stackoverflow.com/questions/2332741/what-is-the-theoretical-maximum-number-of-open-tcp-connections-that-a-modern-lin.

> Each TCP/IP packet has basically four fields for addressing; these are:  
> [source_ip][source_port][destination_ip][destination_port]
> Inside the TCP stack, these four fields are used as a compound key to match up packets to connections (e.g. file descriptors).  
> If a client has many connections to the same port on the same destination, then three of those fields will be the same - only source_port varies to differentiate the different connections. **Ports are 16-bit numbers**, therefore the maximum number of connections any given client can have to any given host port is 64K.

and this page: http://www.freesoft.org/CIE/Course/Section4/8.htm

pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

garrettmoon:
Thank you again for fixing this!

PR #355, opened 2017-05-09
Add request configuration block to allow customizing HTTP request

zachwaugh:
Replaces #321. I had a similar need to #211, where I want to customize request headers on a per request basis, so I added a request configuration block on PINRemoteImageManager as suggested in that thread. My specific use-case is I want to use PINRemoteImageManager to fetch authenticated images for me, which requires setting an Authorization header with our token. However, I don't want to send this for images to other domains, so I need it to be per request. This seemed like the most straightforward approach, but let me know if you think it should be implemented another way.

### DIFF
--- a/Source/Classes/PINRemoteImageManager.m
+++ b/Source/Classes/PINRemoteImageManager.m
@@ -287,7 +287,7 @@ static dispatch_once_t sharedDispatchToken;
     dispatch_async(dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_DEFAULT, 0), ^{
         typeof(self) strongSelf = weakSelf;
         [strongSelf lock];
-        strongSelf.requestConfigurationHandler = configurationBlock;
+            strongSelf.requestConfigurationHandler = configurationBlock;
         [strongSelf unlock];
     });
 }
--- a/Source/Classes/PINRemoteImageManager.m
+++ b/Source/Classes/PINRemoteImageManager.m
@@ -121,6 +121,7 @@ typedef void (^PINRemoteImageManagerDataCompletion)(NSData *data, NSError *error
 @property (nonatomic, assign) float lowQualityBPSThreshold;
 @property (nonatomic, assign) BOOL shouldUpgradeLowQualityImages;
 @property (nonatomic, copy) PINRemoteImageManagerAuthenticationChallenge authenticationChallengeHandler;
+@property (nonatomic, copy) PINRemoteImageManagerRequestConfigurationHandler requestConfigurationHandler;
 @property (nonatomic, strong) NSMutableDictionary <NSString *, NSString *> *httpHeaderFields;
 #if DEBUG
 @property (nonatomic, assign) float currentBPS;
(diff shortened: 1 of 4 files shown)


pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

zachwaugh:
@garrettmoon you mentioned in the initial PR about locking access to the `requestConfigurationHandler ` instance var. I tried that, but it was causing a deadlock. I'm no expert on locking, but I'm assuming because this method is called from `sessionTaskWithURL:...` which itself is called within a lock - https://github.com/pinterest/PINRemoteImage/blob/master/Source/Classes/PINRemoteImageManager.m#L747-L757

Source/Classes/PINRemoteImageManager.m
```diff
@@ -281,6 +282,16 @@ - (void)setValue:(nullable NSString *)value forHTTPHeaderField:(nullable NSStrin
     });
 }
 
+- (void)setRequestConfiguration:(PINRemoteImageManagerRequestConfigurationHandler)configurationBlock {
```
garrettmoon: @zachwaugh you're right about the deadlock.

Here's what I would suggest: We need to add a getter which holds the lock when it accesses `requestConfigurationHandler` or else it isn't thread safe to access by clients. However, since we know we have the lock in downloadDataWithURL: we can access the ivar directly instead of using the property.

Source/Classes/PINRemoteImageManager.m
```diff
@@ -921,6 +932,10 @@ - (NSURLSessionDataTask *)downloadDataWithURL:(NSURL *)url
         request.allHTTPHeaderFields = headers;
     }
     
+    if (self.requestConfigurationHandler) {
+        request = [self.requestConfigurationHandler(request) mutableCopy];
```
garrettmoon: So this should be:

```
if (_requestConfigurationHandler) {
    request = [_requestConfigurationHandler(request) mutableCopy];
}
```

zachwaugh: Thanks, just pushed that change.

Source/Classes/PINRemoteImageManager.m
zachwaugh: @garrettmoon sorry for the confusion, but are you saying we *do* need to a getter, or we don't since we can just use the direct ivar access in `downloadDataWithURL:...`. I made that direct ivar change and it's the only place this variable is accessed currently.

pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

Source/Classes/PINRemoteImageManager.m
garrettmoon: I think we *do* need a getter because the property is exposed publicly? Reading the property externally is currently not thread safe unless I'm mistaken.

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/e21ab974b7a41f173b55a4633ca3a73a/raw/e6d0dbf41483379abb69dcca61fef39b8e1ab68f/log.txt)

pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

Source/Classes/PINRemoteImageManager.m
zachwaugh: The `requestConfigurationHandler` property isn't exposed publicly, only the `setRequestConfiguration:` method.

garrettmoon: Sheesh, I'm doing a terrible job reading this code for some reason. 🤦‍♂️

garrettmoon:
@zachwaugh Thanks so much for adding this and addressing my misguided feedback!

zachwaugh:
@garrettmoon no problem!

PR #356, opened 2017-05-10
Put in more asserts to fix iOS 9 crash :/ #trivial

garrettmoon:
Finally tracked down an iOS 9 device and was able to reproduce the two cases we've been seeing. In both instances the OS is calling delegate methods after completion so it's definitely an OS level bug. This should work around it.

### DIFF
--- a/Source/Classes/PINRemoteImageManager.m
+++ b/Source/Classes/PINRemoteImageManager.m
@@ -618,11 +618,11 @@ static dispatch_once_t sharedDispatchToken;
                               //Remove completion and try again
                               typeof(self) strongSelf = weakSelf;
                               [strongSelf lock];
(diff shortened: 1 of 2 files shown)


pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

timonus:
Do you think it's safer to drop the call if there's no queue or to dispatch it to some other queue?

garrettmoon:
@timonus I think it's safer to drop the call. The delegate methods are currently guaranteed to be processed in order (the queue is serial) and calling them out of order would propagate this bug further down the chain :(

PR #374, opened 2017-06-15
Fix deadlock when canceling processor task

### LINKED ISSUE #372: Images fail to download when using processor
I noticed a regression on master and tracked it down to this commit https://github.com/pinterest/PINRemoteImage/commit/3781e4bc39466412ac745770eb4533fe90e7bf77. When downloading a bunch of images using `processorKey` and `processor`, I don't receive any completion callbacks for the majority of them. The first few requests work, but the rest don't. For the majority that don't work, it appears the network request is never made looking at the server logs. I'm still digging into to try to understand all the changes, but thought I would open the issue in case something jumps out.

zachwaugh:
Use direct ivar access for `_downloadTaskUUID` since `self.downloadTaskUUID` uses a lock as well and causes a deadlock when called here. Fixes #372

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 ## master
 * Add your own contributions to the next release on the line below this with your name.
+- [fixed] Fixes a deadlock with canceling processor tasks [#374](https://github.com/pinterest/PINRemoteImage/pull/374) [zachwaugh](https://github.com/zachwaugh)
 - [fixed] Fixes a deadlock in the retry system. [garrettmoon](https://github.com/garrettmoon)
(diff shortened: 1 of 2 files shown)


pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

garrettmoon:
@zachwaugh Thank you so much for the investigation and fix here!

@maicki my opinion is it's not necessary here since it would essentially just be a method that purely wrapped the ivar?

garrettmoon:
@zachwaugh Mind adding an entry to the CHANGELOG?

zachwaugh:
@garrettmoon sure thing, just updated it

PR #394, opened 2017-08-15
[PINRemoteImageManager] Use the good old NSTemporaryDirectory

nguyenhuy:
The method call to `-[NSFileManager temporaryDirectory]` was added in [#391](https://github.com/pinterest/PINRemoteImage/pull/391/files#diff-9cae23e914261e3df9a01ab2a607b263R234). However, it is only available on iOS 10.0+ and MacOS 10.12+ ([source](https://developer.apple.com/documentation/foundation/nsfilemanager/1642996-temporarydirectory?language=objc)), thus causing crashes on iOS 9. 

Fix by switching to `NSTemporaryDirectory()`. Although it's not the preferred way (see the discussion [here](https://developer.apple.com/documentation/foundation/1409211-nstemporarydirectory)), it's good enough for now.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 * Add your own contributions to the next release on the line below this with your name.
 - [new] Added a way to specify custom retry logic when network error happens [#386](https://github.com/pinterest/PINRemoteImage/pull/386)
 
-- [new] Improve disk cache migration performance [#391](https://github.com/pinterest/PINRemoteImage/pull/391) [chuganzy](https://github.com/chuganzy)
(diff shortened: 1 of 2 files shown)


pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/c1e90130a749e9fb4e1c630128208d8b/raw/7ff6078a64836ca490f9fac6f87ea0a8fd1ad9d4/log.txt)

nguyenhuy:
Merging, thanks @maicki!

chuganzy:
Good catch! Thanks a lot 🙇

PR #420, opened 2017-10-23
Performance tweaks for animated images

garrettmoon:


pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

garrettmoon:
@Adlai-Holler I actually looked into using NSCache, but if we did that we wouldn't be able to shrink and expand our render ranges based on the frequency of the memory warning. In addition, we need to keep track of what is and isn't in the cache at any time so we'd have to monitor the cache with delegate methods. Overall seemed more trouble than keeping a dictionary ourselves.

Source/Classes/AnimatedImages/PINCachedAnimatedImage.m
```diff
@@ -97,9 +98,6 @@ - (instancetype)initWithAnimatedImage:(id <PINAnimatedImage>)animatedImage
         // dispatch later so that blocks can be set after init this runloop
         dispatch_async(dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_DEFAULT, 0), ^{
             [self imageAtIndex:0];
```
appleguy: @garrettmoon shouldn't this line call -coverImage, which seems like it will load image at index 0 if needed, but also will trigger the coverImageReady callback like it did before this change?

Source/Classes/AnimatedImages/PINCachedAnimatedImage.m
```diff
@@ -351,10 +379,12 @@ - (CFTimeInterval)durationAtIndex:(NSUInteger)index
 
 - (BOOL)playbackReady
 {
-    __block BOOL playbackReady = NO;
-    [_lock lockWithBlock:^{
-        playbackReady = _playbackReady == 0;
-    }];
+    static BOOL playbackReady = NO;
```
appleguy: @garrettmoon I think this might be a bug - the static will be the same variable for all instances, but this needs to be different per instance right?

Source/Classes/AnimatedImages/PINCachedAnimatedImage.m
```diff
@@ -395,13 +425,16 @@ - (void)setCoverImageReadyCallback:(PINAnimatedImageInfoReady)coverImageReadyCal
  */
 - (void)clearAnimatedImageCache
 {
-    [_lock lockWithBlock:^{
-        _coverImage = nil;
-        [_cachedOrCachingFrames enumerateIndexesUsingBlock:^(NSUInteger idx, BOOL * _Nonnull stop) {
-            [_frameCache removeObjectForKey:@(idx)];
+    dispatch_async(dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_DEFAULT, 0), ^{
```
appleguy: It would be ideal if these dispatches first checked if on the main queue, and if not, ran synchronously - in general that would provide a bit more control & less overhead for apps that want to drive these calls.

Source/Classes/AnimatedImages/PINCachedAnimatedImage.m
garrettmoon: Man… such a dumb mistake. Will put a fix up.

PR #421, opened 2017-10-23
Enable PINCache for OSX in Podspec

docterd:


garrettmoon:
Thank you for fixing!

PR #432, opened 2017-12-04
Update tvOS example to Swift 4

idrougge:
The tvOS example was never migrated from Swift 2 and thus failed to compile or even auto-convert in Xcode 9. Workspace has been updated to Swift 3.2/4.


garrettmoon:
@idrougge thank you so much!

PR #433, opened 2017-12-20
Fixes a dumb static var which should have been an intsance var. #trivial

garrettmoon:
Also a bit of cleanup.

### DIFF
--- a/Source/Classes/AnimatedImages/PINCachedAnimatedImage.m
+++ b/Source/Classes/AnimatedImages/PINCachedAnimatedImage.m
@@ -36,7 +36,8 @@ static const CFTimeInterval kSecondsBetweenMemoryWarnings = 15;
     PINAnimatedImageInfoReady _coverImageReadyCallback;
     dispatch_block_t _playbackReadyCallback;
     NSMutableDictionary *_frameCache;
-    NSInteger _playbackReady; // Number of frames to cache until playback is ready.
+    NSInteger _frameRenderCount; // Number of frames to cache until playback is ready.
+    BOOL _playbackReady;
     PINOperationQueue *_operationQueue;
     dispatch_queue_t _cachingQueue;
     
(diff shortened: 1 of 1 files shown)


Source/Classes/AnimatedImages/PINCachedAnimatedImage.m
```diff
@@ -302,7 +305,7 @@ - (void)l_cacheFrame:(NSUInteger)frameIndex
     if ([_cachedOrCachingFrames containsIndex:frameIndex] == NO && _cacheCleared == NO) {
         PINLog(@"Requesting: %lu", (unsigned long)frameIndex);
         [_cachedOrCachingFrames addIndex:frameIndex];
-        _playbackReady++;
+        _frameRenderCount++;
```
appleguy: Maybe we could set _playbackReady here "proactively", instead of leaving it NO until access to the getter method causes it to check _frameRenderCount?

maicki: Agree with @appleguy to we could set  the `_playbackReady` in here, that said I don't think we will win anything in here, in fact I would said having fewer state changes in the code base is better e.g. if we can derive state from other variables like in this place the `playbackReady` state will result in fewer bugs and maintenance costs later.

Source/Classes/AnimatedImages/PINCachedAnimatedImage.m
```diff
@@ -379,12 +382,13 @@ - (CFTimeInterval)durationAtIndex:(NSUInteger)index
...
 {
-    static BOOL playbackReady = NO;
-    if (playbackReady == NO) {
-        [_lock lockWithBlock:^{
-            playbackReady = _playbackReady == 0;
-        }];
-    }
+    __block BOOL playbackReady = NO;
+    [_lock lockWithBlock:^{
+        if (_playbackReady == NO) {
```
maicki: Do we need to have the `_playbackReady` instance variable at all? what about just having:

```
- (BOOL)playbackReady
    __block BOOL playbackReady = NO;
    [_lock lockWithBlock:^{
       playbackReady = (_frameRenderCount == 0);
    }];
    return playbackReady;
}
```

From a cost perspective should be even cheaper as we just check the `_frameRenderCount` variable and not have to access the `_playbackReady` as well as the `_frameRenderCount` eventually? Just a thought though.

Issue #87, opened 2015-10-28
Add FLAnimatedImage as private carthage dependency


bdorfman:
Once they merge in this PR https://github.com/Flipboard/FLAnimatedImage/pull/95 FLAnimatedImage should be added as a private dependency in Cartfile.private so that pre-compiled binaries of PINRemoteImage can have animated image support built in.

chuganzy:
+1

PR #441, opened 2018-03-21
Add FLAnimatedImage as Carthage dependency

chuganzy:
Try to close https://github.com/pinterest/PINRemoteImage/issues/87

Or is there any other ways to enable GIF animation while using Carthage??


garrettmoon:
@chuganzy this will add FLAnimatedImage for all Carthage users. I'm not a Carthage expert, do you know if there's a way to add in optional support for FLAnimatedImage?

chuganzy:
@garrettmoon Ok, so you want to make it optional.. Then, there is no way other than checking if it's linked on runtime by checking `FLAnimatedImage` class existence.

chuganzy:
Also making the new target works.

garrettmoon:
@chuganzy That's a good point! I think a separate FLAnimatedImage target might be the way to go?

chuganzy:
## Weakly link `FLAnimatedImage`
### Pros
* Can make `FLAnimatedImage` optional to build
* Can use the existing framework target

### Cons
* Need to check `[FLAnimatedImage class]` existence in some place on runtime
* Need to check `FLAnimatedImage` API existence on runtime

## Make new target
### Pros
* Nothing to check on runtime

### Cons
* Force users to build `FLAnimatedImage` even if they do not need it

garrettmoon:
@chuganzy I think we're actually going to be removing support for FLAnimatedImage in favor of native GIF processing. Would you all be willing to give it a try? You should be able to use PINAnimatedImageView instead of FLAnimatedImageView. Would love to any feedback! Feel free to send it to garrett@pinterest.com

garrettmoon:
Here's the PR that added PINAnimatedImageView if it helps: https://github.com/pinterest/PINRemoteImage/pull/453

chuganzy:
Aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaawsome!!

chuganzy:
Closing since PINRemoteImage no longer uses FLAnimatedImage.

garrettmoon:
@chuganzy please let me know if you all run into any issues!

PR #442, opened 2018-03-21
Fix warnings related to block declaration and referencing self within a block

davamale:
## Fix warnings related to block declaration and referencing self within a block

 - fixes retain cycle warning (calling self inside a block)
- fix warning in block declaration


davamale:
Ok, will do.

Thanks

garrettmoon:
Hey @davamale I think there may be something wrong with this PR, it looks like it's including some old changes. Can you try cherry-picking the changes on to a new branch off master?

I also think these may already be fixed in master?

davamale:
Yes, I'm dropping this PR.

Thanks though

Issue #451, opened 2018-04-19
Potential bug with PINRemoteImageBasicCache causing decoding of images on the main thread


kgaidis:
I think there's a bug where images are decoded on the main thread when they are first downloaded IF `PINRemoteImageBasicCache` is used (AKA no PINCache).

When an image is downloaded, or pulled from cache, everything trickles down into `PINRemoteImageManager` method:
```
- (BOOL)materializeAndCacheObject:(id)object
                      cacheInDisk:(NSData *)diskData
                   additionalCost:(NSUInteger)additionalCost
                              url:(NSURL *)url
                              key:(NSString *)key
                          options:(PINRemoteImageManagerDownloadOptions)options
                         outImage:(PINImage **)outImage
                        outAltRep:(id *)outAlternateRepresentation
```

Also, when an image is FIRST downloaded, `object` is NSData, and `cacheInDisk` is also NSData.

The `materializeAndCacheObject:` keeps going and eventually saves `PINRemoteImageMemoryContainer` to memory:
```
if (updateMemoryCache) {
    [container.lock lockWithBlock:^{
        NSUInteger cacheCost = additionalCost;
        cacheCost += [container.data length];
        CGImageRef imageRef = container.image.CGImage;
        NSAssert(container.image == nil || imageRef != NULL, @"We only cache a decompressed image if we decompressed it ourselves. In that case, it should be backed by a CGImageRef.");
        if (imageRef) {
            cacheCost += CGImageGetHeight(imageRef) * CGImageGetBytesPerRow(imageRef);
        }
        [self.cache setObjectInMemory:container forKey:key withCost:cacheCost];
    }];
}
```

`PINRemoteImageBasicCache` implementation for `setObjectInMemory:`:
```
-(void)setObjectInMemory:(id)object forKey:(NSString *)key withCost:(NSUInteger)cost
{
    [self.cache setObject:object forKey:key cost:cost];
}
```

That all is great. 

Then the `materializeAndCacheObject:` keeps going and eventually saves `cacheInDisk` to "disk:"
```
if (diskData) {
    [self.cache setObjectOnDisk:diskData forKey:key];
}
```

`PINRemoteImageBasicCache` implementation for `setObjectOnDisk:`:
```
-(void)setObjectOnDisk:(id)object forKey:(NSString *)key
{
    [self.cache setObject:object forKey:key];
}
```

At that point, the previously saved `PINRemoteImageMemoryContainer` gets overriden by `NSData`.

[...]

PR #457, opened 2018-05-03
Fixed a bug where using PINRemoteImageBasicCache would cause images to be decoded on the main thread.

kgaidis:
Issue reference: #451 

-  PINRemoteImageBasicCache disk-cache methods are now no-op.
- Changed `cancelTaskWithUUID:` documentation to say disk cache, and not memory cache.

Feel free to (of course) to pick on anything that seems weird. One decision that may feel odd is that I left the global-queue code in `objectFromDiskForKey:completion:`. My thought-process was to leave the code functioning as it did before as to not cause any unknown side effects. In this case, the completionHandler will continue to return from a global queue (as before), but it will just _always_ return a nil for `object`, instead of just _sometimes_.

### DIFF
--- a/Source/Classes/PINRemoteImageBasicCache.h
+++ b/Source/Classes/PINRemoteImageBasicCache.h
@@ -11,7 +11,8 @@
 
 /**
  *  Simplistic <PINRemoteImageCacheProtocol> wrapper based on NSCache.
- *  not persisting any data on disk
+ *
+ *  No data is persisted on disk. The disk cache methods are no-op.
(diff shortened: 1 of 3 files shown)


pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

PR #460, opened 2018-05-24
Use non-nil session configuration

### LINKED ISSUE #459: Calling [NSURLSession sessionWithConfiguration:] with nil configuration object.
https://github.com/pinterest/PINRemoteImage/blob/ca736620572dc2d1f9f7a873c0ec6cf59bfcb02c/Source/Classes/PINRemoteImageManager.m#L188

"configuration" in above line could be nil when called from [PINRemoteImageManager sharedImageManager]. Prior to 3.0.0-beta.11, "configuration" is checked against nil and a new [NSURLSessionConfiguration ephemeralSessionConfiguration] object is created if it is nil few lines above.

But a bug got introduced in 3.0.0-beta.11 where property _sessionConfiguration is checked against nil and new object is created and stored in _sessionConfiguration but original "configuration" parameter is still passed to above function essentially calling [NSURLSession sessionWithConfiguration:] with nil configuration object.

garrettmoon:
In a recent refactor a bug was introduced where the session configuration that was
created if you passed in nil was not being used in initializing the session manager.

Fixes #459, thank you for reporting @nagasunilt!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 - [fixed] Fixes a bug where using PINRemoteImageBasicCache would cause images to be decoded on the main thread. [#457](https://github.com/pinterest/PINRemoteImage/pull/457) [kgaidis](https://github.com/kgaidis)
 - [cleanup] Remove unused code that supported iOS < 7. [#435](https://github.com/pinterest/PINRemoteImage/pull/435) [Adlai-Holler](https://github.com/Adlai-Holler)
 - [cleanup] Use NS_ERROR_ENUM to improve Swift import. [#440](https://github.com/pinterest/PINRemoteImage/pull/440) [Adlai-Holler](https://github.com/Adlai-Holler)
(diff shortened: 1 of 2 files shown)


pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

naga-cf:
Thanks for the quick turnaround!

nagasunilt:
Thanks for the quick turnaround!

quintonparker:
Great turnaround indeed. We (spree.co.za) are users of this library. Don't mean to be pushy but when can we expect this in an official release tag?

garrettmoon:
@quintonparker should be in a beta release or RC soon, as far as an non-beta release it may be a while 😬

quintonparker:
@garrettmoon thanks!

PR #462, opened 2018-06-08
Respect Cache-Control headers

wiseoldduck:
The PINCache upgrade bundled into this has also been submitted as a standalone PR

### DIFF
--- a/Source/Classes/PINCache/PINCache+PINRemoteImageCaching.m
+++ b/Source/Classes/PINCache/PINCache+PINRemoteImageCaching.m
@@ -13,12 +13,12 @@
 //******************************************************************************************************
 // Memory cache methods
 //******************************************************************************************************
--(nullable id)objectFromMemoryForKey:(NSString *)key
+- (nullable id)objectFromMemoryForKey:(NSString *)key
 {
     return [self.memoryCache objectForKey:key];
 }
 
--(void)setObjectInMemory:(id)object forKey:(NSString *)key withCost:(NSUInteger)cost
+- (void)setObjectInMemory:(id)object forKey:(NSString *)key withCost:(NSUInteger)cost
 {
     [self.memoryCache setObject:object forKey:key withCost:cost];
 }
@@ -31,12 +31,12 @@
 //******************************************************************************************************
 // Disk cache methods
 //******************************************************************************************************
--(nullable id)objectFromDiskForKey:(NSString *)key
+- (nullable id)objectFromDiskForKey:(NSString *)key
 {
     return [self.diskCache objectForKey:key];
 }
 
--(void)objectFromDiskForKey:(NSString *)key completion:(PINRemoteImageCachingObjectBlock)completion
+- (void)objectFromDiskForKey:(NSString *)key completion:(PINRemoteImageCachingObjectBlock)completion
 {
     __weak typeof(self) weakSelf = self;
     [self.diskCache objectForKeyAsync:key completion:^(PINDiskCache * _Nonnull cache, NSString * _Nonnull key, id<NSCoding>  _Nullable object) {
@@ -47,12 +47,12 @@
     }];
 }
 
--(void)setObjectOnDisk:(id)object forKey:(NSString *)key withAgeLimit:(NSTimeInterval)ageLimit
+- (void)setObjectOnDisk:(id)object forKey:(NSString *)key withAgeLimit:(NSTimeInterval)ageLimit
 {
     [self.diskCache setObject:object forKey:key withAgeLimit:ageLimit];
 }
 
--(void)setObjectOnDisk:(id)object forKey:(NSString *)key
+- (void)setObjectOnDisk:(id)object forKey:(NSString *)key
 {
     [self setObject:object forKey:key withAgeLimit:0];
 }
@@ -76,6 +76,16 @@
   } else {
     [self removeObjectForKeyAsync:key completion:nil];
   }
+}
+
+- (BOOL)diskCacheIsTTLCache
+{
+    return self.diskCache.isTTLCache;
+}
+
+- (BOOL)memoryCacheIsTTLCache
+{
+    return self.memoryCache.isTTLCache;
 }
 
 @end
--- a/Source/Classes/PINCache/PINCache+PINRemoteImageCaching.m
+++ b/Source/Classes/PINCache/PINCache+PINRemoteImageCaching.m
@@ -18,6 +18,11 @@
     return [self.memoryCache objectForKey:key];
 }
 
+- (void)setObjectInMemory:(id)object forKey:(NSString *)key withCost:(NSUInteger)cost withAgeLimit:(NSTimeInterval)ageLimit
+{
+    [self.memoryCache setObject:object forKey:key withCost:cost ageLimit:ageLimit];
+}
+
 - (void)setObjectInMemory:(id)object forKey:(NSString *)key withCost:(NSUInteger)cost
 {
     [self.memoryCache setObject:object forKey:key withCost:cost];
--- a/Source/Classes/PINCache/PINCache+PINRemoteImageCaching.m
+++ b/Source/Classes/PINCache/PINCache+PINRemoteImageCaching.m
@@ -54,7 +54,7 @@
 
 -(void)setObjectOnDisk:(id)object forKey:(NSString *)key
 {
-    [self.diskCache setObject:object forKey:key];
+    [self setObject:object forKey:key withAgeLimit:0];
 }
 
 - (BOOL)objectExistsForKey:(NSString *)key
--- a/Source/Classes/PINCache/PINCache+PINRemoteImageCaching.m
+++ b/Source/Classes/PINCache/PINCache+PINRemoteImageCaching.m
@@ -47,6 +47,11 @@
     }];
 }
 
+-(void)setObjectOnDisk:(id)object forKey:(NSString *)key withAgeLimit:(NSTimeInterval)ageLimit
+{
+    [self.diskCache setObject:object forKey:key withAgeLimit:ageLimit];
+}
+
 -(void)setObjectOnDisk:(id)object forKey:(NSString *)key
 {
     [self.diskCache setObject:object forKey:key];
--- a/Source/Classes/PINRemoteImageManager.h
+++ b/Source/Classes/PINRemoteImageManager.h
@@ -206,10 +206,24 @@ typedef void(^PINRemoteImageManagerMetrics)(NSURL  * __nonnull url, NSURLSession
 
 /**
  The result of this method is assigned to self.cache in init. If you wish to provide a customized cache to the manager you can subclass PINRemoteImageManager and return a custom object, implementing PINRemoteImageCaching protocol from this method. Same effect could be achieved by using initWithSessionConfiguration:alternativeRepresentationProvider:imageCache: initializer.
+ @deprecated Use the class method +defaultImageCache
+ @warning This method is meant only for override. It will be called *once* by an instance of PINRemoteImageManager. The default implementation creates a new cache on every call. If you're looking to access the cache being used by an instance of PINRemoteImageManager, @c cache.
+ @return An instance of a object, implementing PINRemoteImageCaching protocol.
+ */
+- (nonnull id<PINRemoteImageCaching>)defaultImageCache __attribute__((deprecated));
+
+/**
+ The result of this method is assigned to self.cache in init. If you wish to provide a customized cache to the manager you can subclass PINRemoteImageManager and return a custom object, implementing PINRemoteImageCaching protocol from this method. Same effect could be achieved by using initWithSessionConfiguration:alternativeRepresentationProvider:imageCache: initializer.
   @warning This method is meant only for override. It will be called *once* by an instance of PINRemoteImageManager. The default implementation creates a new cache on every call. If you're looking to access the cache being used by an instance of PINRemoteImageManager, @c cache.
  @return An instance of a object, implementing PINRemoteImageCaching protocol.
  */
-- (nonnull id<PINRemoteImageCaching>)defaultImageCache;
++ (nonnull id<PINRemoteImageCaching>)defaultImageCache;
+
+/**
+ * If you want a Ttl (maxAge) image cache, you can pass the result of this method explicitly into the initWithSessionConfiguration:alternativeRepresentationProvider:imageCache: initializer.
+ * @return An instance of a object, implementing PINRemoteImageCaching protocol.
+ */
++ (nonnull id<PINRemoteImageCaching>)defaultImageTtlCache;
 
 /**
  * Sets a custom header to be included in every request. Headers set from this method will override any header from NSURLSessionConfiguration.
--- a/Source/Classes/PINRemoteImageManager.h
+++ b/Source/Classes/PINRemoteImageManager.h
@@ -62,10 +62,12 @@ typedef NS_OPTIONS(NSUInteger, PINRemoteImageManagerDownloadOptions) {
     PINRemoteImageManagerDownloadOptionsSkipRetry = 1 << 5,
     /**
      * Do not honor HTTP Cache-Control headers
-     * Currently PINRemoteImage will by default respect 'no-store', 'no-cache', 'max-age', 'Expires', and 'must-revalidate'. Set this flag to ignore those headers.
-     * TODO: Currently PINRemoteImage will re-download images that only must be re-validated. In the future this could be improved with revalidation behavior that stores ETag or Last-Modified values and only
-     * makes HEAD requests to see if these headers are unchanged.
-     * see https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control
+     * By default, PINRemoteImage will by default respect 'no-store', 'no-cache', 'max-age',
+     * 'Expires', and 'must-revalidate'. Set this flag to ignore those headers.
+     * TODO: Currently PINRemoteImage will re-download images that only must be re-validated. In the
+     * future this could be improved with revalidation behavior that stores ETag or Last-Modified
+     * values and only makes HEAD requests to see if these headers are unchanged.
+     * see https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control and
      * https://tools.ietf.org/html/rfc7234*/
     PINRemoteImageManagerDownloadOptionsIgnoreCacheControlHeaders = 1 << 6
 };
--- a/Source/Classes/PINRemoteImageManager.h
+++ b/Source/Classes/PINRemoteImageManager.h
@@ -60,6 +60,14 @@ typedef NS_OPTIONS(NSUInteger, PINRemoteImageManagerDownloadOptions) {
     PINRemoteImageManagerDownloadOptionsIgnoreCache = 1 << 4,
     /** Skip download retry */
     PINRemoteImageManagerDownloadOptionsSkipRetry = 1 << 5,
+    /**
+     * Do not honor HTTP Cache-Control headers
+     * Currently PINRemoteImage will by default respect 'no-store', 'no-cache', 'max-age', 'Expires', and 'must-revalidate'. Set this flag to ignore those headers.
+     * TODO: Currently PINRemoteImage will re-download images that only must be re-validated. In the future this could be improved with revalidation behavior that stores ETag or Last-Modified values and only
+     * makes HEAD requests to see if these headers are unchanged.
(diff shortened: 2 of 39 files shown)


pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          2 Warnings
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">This is a big PR, please consider splitting it up to ease code review.</td>
    </tr>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">This is a big PR, please consider splitting it up to ease code review.</td>
    </tr>
  </tbody>
</table>

</p>

pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">This is a big PR, please consider splitting it up to ease code review.</td>
    </tr>
  </tbody>
</table>

</p>

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/cd9c5bb94c9717e8baacf266d916a0e5/raw/8ce37d2796de7f58ad740ef29dc834c119504c44/log.txt)

pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">This is a big PR, please consider splitting it up to ease code review.</td>
    </tr>
  </tbody>
</table>

</p>

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/bd9a3d4b446fb07804ae42a1447ca476/raw/f48aa6196cf74b125fb92c89f7fb74f6c51aa7e1/log.txt)

pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">This is a big PR, please consider splitting it up to ease code review.</td>
    </tr>
  </tbody>
</table>

</p>

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/9b392435ccb76881b4075ef7e3569f13/raw/f3ade7bed56b598fff14a07d7453289436890142/log.txt)

pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">This is a big PR, please consider splitting it up to ease code review.</td>
    </tr>
  </tbody>
</table>

</p>

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/28d924df8e680577f41b6d3f991ec2ea/raw/9d7f7d83b3962c9e15c5b2227888e1955e9d48bf/log.txt)

Source/Classes/PINRemoteImageManager.m
```diff
@@ -116,6 +116,7 @@ @interface PINRemoteImageManager () <PINURLSessionManagerDelegate>
 @property (nonatomic, copy) id<PINRequestRetryStrategy> (^retryStrategyCreationBlock)(void);
 @property (nonatomic, copy) PINRemoteImageManagerRequestConfigurationHandler requestConfigurationHandler;
 @property (nonatomic, strong) NSMutableDictionary <NSString *, NSString *> *httpHeaderFields;
+@property(nonatomic, assign) BOOL isTtlCache;
```
appleguy: To match the codebase, space before ( -- which is not done in some other codebases.

I would suggest a different name, maybe respectCacheControlHeaders, trackImageExpiry, enforceCacheAgeLimits - since this class is not by itself a cache, it's more like a class managing a cache than "isTtlCache".

Source/Classes/PINRemoteImageManager.m
```diff
@@ -784,13 +786,62 @@ - (void)downloadImageWithURL:(NSURL *)url
             NSError *remoteImageError = error;
             PINImage *image = nil;
             id alternativeRepresentation = nil;
-             
+            __block NSNumber *maxAge = nil;
```
appleguy: Any reason to box this as an NSNumber? If not, NSInteger should result shorter / faster code, as long as cases like NSNotFound or other sentinel values are handled.

Source/Classes/PINRemoteImageManager.m
```diff
@@ -784,13 +786,62 @@ - (void)downloadImageWithURL:(NSURL *)url
...
             if (remoteImageError == nil) {
-                 //stores the object in the caches
-                 [self materializeAndCacheObject:data cacheInDisk:data additionalCost:0 url:url key:key options:options outImage:&image outAltRep:&alternativeRepresentation];
+                if (_isTtlCache && !(options & PINRemoteImageManagerDownloadOptionsIgnoreCacheControlHeaders)) {
```
appleguy: Believe it or not, this bitwise-AND should not be inverted / handled as a BOOL as it can fail on specifically 32-bit ARM! Our team had some nasty issues with this last Fall.

If other parts of the file have this bug as well, we should introduce a macro similar to ASInterfaceStateIncludes() or  ASCellLayoutModeIncludes() to resolve.

Source/Classes/PINRemoteImageManager.m
```diff
@@ -784,13 +786,62 @@ - (void)downloadImageWithURL:(NSURL *)url
...
+                    // examine Cache-Control headers (https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control)
+                    if ([response isKindOfClass:[NSHTTPURLResponse class]]) {
+                        NSHTTPURLResponse *httpResponse = (NSHTTPURLResponse *) response;
+
+                        [[[httpResponse allHeaderFields][@"Cache-Control"] componentsSeparatedByString:@","] enumerateObjectsUsingBlock:^(id obj, NSUInteger idx, BOOL *stop) {
```
appleguy: Since these are complex lines of code, and we access this twice, local variable?

NSArray *allHeaderFields = ...

Source/Classes/PINRemoteImageManager.m
```diff
@@ -784,13 +786,62 @@ - (void)downloadImageWithURL:(NSURL *)url
...
+                            NSString *trimmed = [[(NSString *) obj stringByTrimmingCharactersInSet:[NSCharacterSet whitespaceCharacterSet]] lowercaseString];
+
+                            if ([trimmed isEqualToString:@"no-store"] || [trimmed isEqualToString:@"must-revalidate"] || [trimmed isEqualToString:@"no-cache"]) {
+                                maxAge = @(0);
+                                *stop = YES;
```
appleguy: I'd suggest we just use for( ... in ...), as it is both faster - NSFastEnumeration and avoids the internal @autorelease pool that is otherwise created around every block invocation - and it also is a bit nicer looking (less block boilerplate with the arguments, easier to write "break";, etc)

This could be, for (NSString *field in [allHeaderFields[@"Cache-Control"] componentsSeparatedByString:@",") { }

Source/Classes/PINRemoteImageManager.m
```diff
@@ -784,13 +786,62 @@ - (void)downloadImageWithURL:(NSURL *)url
...
+                                NSArray<NSString *> *split = [trimmed componentsSeparatedByString:@"max-age="];
+                                if ([split count] == 2) {
+                                    maxAge = @([split[1] integerValue]);
+                                }
+                            }
+                        }];
+
+                        NSString *expires = nil;
+                        // If there is a Cache-Control header with the "max-age" directive in the response, the Expires header is ignored
+                        if (!maxAge && (expires = [httpResponse allHeaderFields][@"Expires"])) {
```
appleguy: Definitely move this expires = assignment up to the declaration...although it can be useful and crafty in some cases, I think there's no reason to put the assignment in the conditional in this case.

Source/Classes/PINRemoteImageManager.m
```diff
@@ -784,13 +786,62 @@ - (void)downloadImageWithURL:(NSURL *)url
...
+                            // https://developer.apple.com/library/archive/qa/qa1480/_index.html
+
+                            static NSDateFormatter *sRFC7231PreferredDateFormatter;
```
appleguy: Assign nil. I'm not sure we should rely on static variables being zero-initialized.

Also, we should move this out to an accessor method. Ideally static variables are declared outside methods (even if just above the method). In this case a relatively complex method would be shortened if we moved the cached formatter to an accessor like +preferredCacheControlDateFormatter.

Source/Classes/PINRemoteImageManager.m
```diff
@@ -784,13 +786,62 @@ - (void)downloadImageWithURL:(NSURL *)url
...
+                            }
+
+                            date = [sRFC7231PreferredDateFormatter dateFromString:expires];
+
+                            // Invalid dates (notably "0") or dates in the past must not be cached (RFC7231 5.3)
+                            maxAge = @((NSInteger) MAX(([date timeIntervalSinceNow]), 0));
+                        }
+                    }
+                }
+                //stores the object in the caches
```
appleguy: Space & caps after //

Source/Classes/PINRemoteImageManager.m
```diff
@@ -1356,7 +1419,15 @@ - (BOOL)materializeAndCacheObject:(id)object
     }
     
     if (diskData) {
-        [self.cache setObjectOnDisk:diskData forKey:key];
+        // maxAge of 0 is not stored at all
```
appleguy: Could you clarify this comment? Does this mean "maxAge of 0 means that images should not be stored at all" ?

Is there a value for maxAge that means infinite?

Source/Classes/PINRemoteImageManager.h
```diff
@@ -60,6 +60,14 @@ typedef NS_OPTIONS(NSUInteger, PINRemoteImageManagerDownloadOptions) {
     PINRemoteImageManagerDownloadOptionsIgnoreCache = 1 << 4,
     /** Skip download retry */
     PINRemoteImageManagerDownloadOptionsSkipRetry = 1 << 5,
+    /**
+     * Do not honor HTTP Cache-Control headers
+     * Currently PINRemoteImage will by default respect 'no-store', 'no-cache', 'max-age', 'Expires', and 'must-revalidate'. Set this flag to ignore those headers.
+     * TODO: Currently PINRemoteImage will re-download images that only must be re-validated. In the future this could be improved with revalidation behavior that stores ETag or Last-Modified values and only
```
appleguy: Minor / optional: "By default, PINRemoteImage will respect...".

Could also drop "Currently" in the next line, as presumably that will be updated if it becomes supported in the future. The next line also seems like it could use a line break to stay within a ~120 (or a bit more) column width — which isn't a hard rule for these frameworks, but up to ~150 works pretty well for Obj-C.

Great comment overall, thanks for including the IETF and Mozilla links!

Tests/PINRemoteImageTests.m
```diff
@@ -943,6 +944,51 @@ - (void)testQOS
     XCTAssert(dispatch_semaphore_wait(semaphore, [self timeout]) == 0, @"Semaphore timed out.");
 }
 
+- (void)testMaxAge
```
appleguy: Thanks for adding tests! Maybe call this testCacheControlMaxAgeExpiry, or just testCacheControl? Since the maxAge comes from cache control, the cache-control seems like the root nature / focus of the test.

Source/Classes/PINRemoteImageManager.m
wiseoldduck: Yes, we need to distinguish between "set to 0" and "not set" (nil)

Source/Classes/PINRemoteImageManager.m
wiseoldduck: Spacing added.
Since the underlying cache is not necessarily PINCache, this var is storing (caching) whether the cache implements the optional `- (void)setObjectOnDisk:(id)object forKey:(NSString *)key withAgeLimit:(NSTimeInterval)ageLimit` method, that is whether it is a Ttl cache or not. So I think the name is accurate, (although maybe could be `cacheIsTtlCache`)...

Source/Classes/PINRemoteImageManager.m
wiseoldduck: I was mimicking the style used in `-[PINRemoteImageDownloadTask didReceiveResponse:]`, although it is far enough away that maybe we don't care (or maybe want to update *that* to use separate assignments)

Source/Classes/PINRemoteImageManager.m
wiseoldduck: Done (yes that's what it meant).
There is no header that can be sent to indicate "infinite", which is why you will see so many HTTP responses with values like "one year in the future". However not setting a value at all, which in our case is represented by `maxAge == nil`, does effectively mean that. (In our case it will use the cache-global maxAge)

Tests/PINRemoteImageTests.m
wiseoldduck: "testCacheControlSupport"?

Source/Classes/PINRemoteImageManager.m
```diff
@@ -1356,7 +1429,17 @@ - (BOOL)materializeAndCacheObject:(id)object
     }
     
     if (diskData) {
-        [self.cache setObjectOnDisk:diskData forKey:key];
+        // maxAge of 0 means that images should not be stored at all.
+        // There is no HTTP header that can be sent to indicate "infinite". However not setting a value at all, which in
+        // our case is represented by maxAge == nil, effectively means that.
+        if (!(maxAge && [maxAge integerValue] == 0)) {
```
garrettmoon: Is this more readable (honest question)? (I think it's still correct)

```
if (maxAge && [maxAge integerValue] != 0 && _isTtlCache) {
    [self.cache setObjectOnDisk:diskData forKey:key withAgeLimit:[maxAge integerValue]];
} else if (!maxAge) {
    // unset (nil) maxAge, or a cache that is not _isTtlCache behave as before (will use cache global ageLimit)
    [self.cache setObjectOnDisk:diskData forKey:key];
}
```

Tests/PINRemoteImageTests.m
```diff
@@ -943,6 +944,51 @@ - (void)testQOS
     XCTAssert(dispatch_semaphore_wait(semaphore, [self timeout]) == 0, @"Semaphore timed out.");
 }
 
+- (void)testCacheControlSupport
```
garrettmoon: I think these would be (slightly) better tests if they actually checked the results of a second download instead of just seeing if the objects are still in the cache. Or perhaps both? PINRemoteImage will tell you how it got an object in result.resultType.

Source/Classes/PINRemoteImageManager.m
```diff
@@ -116,6 +116,7 @@ @interface PINRemoteImageManager () <PINURLSessionManagerDelegate>
 @property (nonatomic, copy) id<PINRequestRetryStrategy> (^retryStrategyCreationBlock)(void);
 @property (nonatomic, copy) PINRemoteImageManagerRequestConfigurationHandler requestConfigurationHandler;
 @property (nonatomic, strong) NSMutableDictionary <NSString *, NSString *> *httpHeaderFields;
+@property (nonatomic, assign) BOOL isTtlCache;
```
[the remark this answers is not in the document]
wiseoldduck: Yes

Source/Classes/PINRemoteImageManager.m
```diff
@@ -784,13 +805,53 @@ - (void)downloadImageWithURL:(NSURL *)url
...
+                                NSArray<NSString *> *split = [trimmed componentsSeparatedByString:@"max-age="];
+                                if ([split count] == 2)
+                                {
+                                    // if the max-age provided is invalid (does not parse into an
+                                    // int), we wind up with 0 which will be treated as do-not-cache.
+                                    // This is the RFC defined behavior for a malformed "expires" header,
+                                    // and while I cannot find any explicit instruction of how to behave
+                                    // with a malformed "max-age" header, it seems like a reasonable approach.
+                                    maxAge = @([split[1] integerValue]);
+                                }
```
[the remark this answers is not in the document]
wiseoldduck: hmm well the `else` would be if `[split count] == 0`, in which case it was not set at all (and `nil` is correct), or `[split count] > 2`, in which case you are correct but it's a pretty weird case with a header like `Cache-Control: maxage=maxage=444`

Source/Classes/PINCache/PINCache+PINRemoteImageCaching.m
```diff
@@ -47,6 +47,11 @@ -(void)objectFromDiskForKey:(NSString *)key completion:(PINRemoteImageCachingObj
...
 }
 
+-(void)setObjectOnDisk:(id)object forKey:(NSString *)key withAgeLimit:(NSTimeInterval)ageLimit
+{
+    [self.diskCache setObject:object forKey:key withAgeLimit:ageLimit];
+}
+
 -(void)setObjectOnDisk:(id)object forKey:(NSString *)key
 {
     [self.diskCache setObject:object forKey:key];
```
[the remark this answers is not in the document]
wiseoldduck: OK

Source/Classes/PINRemoteImageManager.m
wiseoldduck: I dunno? Yes it looks still correct, and the first case is more readable, but it seems harder to reason through the other two possibilities...

wiseoldduck: No I think it is also incorrect in that if `maxAge` is set but `_isTtlCache` is false, the proposal fails to call the old method (does not cache at all)

Tests/PINRemoteImageTests.m
wiseoldduck: OK.

Well it would be second & third downloads, to confirm that it got it from the cache and then did not.

Source/Classes/PINRemoteImageManager.m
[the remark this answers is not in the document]
wiseoldduck: I added a condition for the pretty weird case

Tests/PINRemoteImageTests.m
wiseoldduck: This change will make more obvious another limitation - The PINCache TTL implementation only applies to the disk cache. If you kept an app alive long enough, without memory warnings, you could still get results from the memory cache that the header would have suggested you shouldn't.

wiseoldduck: Done

Source/Classes/PINRemoteImageManager.m
appleguy: I think this may suggest we should have a simple local variable with a more descriptive name, like:

BOOL neverStoreImage = (maxAge && [maxAge integerValue] == 0);
if (neverStoreImage == NO) {

}

There's probably a better variable name or another way to do it that is more readable still, like you could check:

BOOL storeImageIndefinitely = !maxAge;
if (storeImageIndefinitely || [maxAge integerValue] > 0) {

}

This second one might be the most readable?

Source/Classes/PINRemoteImageManager.m
```diff
@@ -784,13 +805,53 @@ - (void)downloadImageWithURL:(NSURL *)url
             NSError *remoteImageError = error;
             PINImage *image = nil;
             id alternativeRepresentation = nil;
-             
+            __block NSNumber *maxAge = nil;
             if (remoteImageError == nil) {
-                 //stores the object in the caches
-                 [self materializeAndCacheObject:data cacheInDisk:data additionalCost:0 url:url key:key options:options outImage:&image outAltRep:&alternativeRepresentation];
+                BOOL ignoreHeaders = (options & PINRemoteImageManagerDownloadOptionsIgnoreCacheControlHeaders) != 0;
```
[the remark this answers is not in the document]
wiseoldduck: 🤔

Source/Classes/PINRemoteImageManager.m
wiseoldduck: Excellent suggestion! See the newest edition 👍

garrettmoon:
Thanks so much for your contribution @wiseoldduck !

PR #467, opened 2018-06-29
#trivial Move implementation of max-age header scanning into a category.

wiseoldduck:
(On NSHTTPURLResponse)


maicki:
@wiseoldduck Thanks for addressing the comments, let's land it!

PR #475, opened 2018-09-18
Add support for seeking in animated images

joshbeal:
See [Texture RFC](https://github.com/TextureGroup/Texture/issues/1129) for context on this proposed feature


Source/Classes/AnimatedImages/PINAnimatedImageView.h
```diff
@@ -23,4 +23,9 @@
 @property (nullable, nonatomic, strong) NSString *animatedImageRunLoopMode;
 @property (nonatomic, assign, getter=isPlaybackPaused) BOOL playbackPaused;
 
+/**
+ @abstract Seek to a play head position of an animated image.
```
garrettmoon: Can we add to the comment and mention that this may not be performant? Because the GIF is currently memory backed and caches ahead to optimize for performance.

PR #477, opened 2018-09-20
Allow clients to configure URLSession URLCache if they choose.

wiseoldduck:
Set NSURLRequestReloadIgnoringLocalCacheData in the default  session instead of per-request. This lets a client configure a session with a URLCache if they so choose. Also explicitly set URLCache in the default session to  nil to avoid the small costs associated with the default ephemeral session memory cache that is never used.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## master
 * Add your own contributions to the next release on the line below this with your name.
-- [new] Respect Cache-Control and Expires headers if the cache supports TTL [#462](https://github.com/pinterest/PINRemoteImage/pull/462) [wiseoldduck](https://github.com/wiseoldduck)
+- [new] Allow use of NSURLCache via a custom NSURLSession [#477](https://github.com/pinterest/PINRemoteImage/pull/477) [wiseoldduck](https://github.com/wiseoldduck)
+- [new] Respect Cache-Control and Expires headers if the cache supports TTL. [#462](https://github.com/pinterest/PINRemoteImage/pull/462) [wiseoldduck](https://github.com/wiseoldduck)
 - [new] Updated to latest PINCache beta 7. [wiseoldduck](https://github.com/wiseoldduck)
 - [iOS11] Fix warnings [#428](https://github.com/pinterest/PINRemoteImage/pull/428) [Eke](https://github.com/Eke)
 - [new / beta] Native Support for GIFs and animated WebP [#453](https://github.com/pinterest/PINRemoteImage/pull/453) [garrettmoon](https://github.com/garrettmoon)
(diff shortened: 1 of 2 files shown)


pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

rainypixels:
@wiseoldduck I was poking around in PINRemoteImage/PINCache to see if they respect cache-control headers, and I happened upon this pull request and #462. 🎉I have a couple of quick questions for you to ensure I’m thinking about all this correctly.

In our app, we have images that continue to live at the same url; even when the file changes, the url doesn’t. Effectively, we need clients to periodically refresh the images they have downloaded. The support for cache-control headers that you built into #462 is awesome, but as you mention in the comments, it is incomplete ([revalidation behavior is a todo](https://github.com/pinterest/PINRemoteImage/blob/master/Source/Classes/PINRemoteImageManager.h#L66)). So, we are wondering if we should be using a `URLCache` to get a more complete solution to caching.

Presumably, I need to init with a custom `URLSessionConfiguration`? Are there any downsides to this? We still want all the decompression, image caching, etc. benefits we get between PINRemoteImage and PINCache, and I'm assuming we still will continue to get those, right?

If there is no downside to using a custom `URLSessionConfiguration`, I’m curious why you decided to implement cache-control header support in PINRemoteImage in the first place. I'm sure there was a good reason for it.

Thanks for this great pull request (and of course to @garrettmoon and crew for PINRemoteImage which came into my life through Texture)!

wiseoldduck:
We were expecting the PINCache w/TTL to perform better than it did. That was the reason for the internal cache-control implementation. `NSURLCache` is also more opaque if it doesn't happen to behave exactly like you want. But I would indeed probably set a `NSURLSessionConfiguration.URLCache` and not use `PINCache` if I was interested in easy standards-compliant cache behavior.

However the case you describe _is_ within what the PRI implementation supports - it _over_-requests images, not under, downloading them again when it only needed to check the headers and _maybe_ re-download them. So that's unoptimized, but not such that you would get the incorrect image in the end.

rainypixels:
> We were expecting the PINCache w/TTL to perform better than it did. That was the reason for the internal cache-control implementation.

Ah, that makes sense!

> But I would indeed probably set a NSURLSessionConfiguration.URLCache and not use PINCache if I was interested in easy standards-compliant cache behavior.

So if I'm reading this correctly, if use `NSURLSessionConfiguration.URLCache`, I will lose benefits like [PINCache processing the responses into decompressed `UIImage`s](https://github.com/pinterest/PINCache/issues/95#issuecomment-233014211)? If that's the case, I'd rather crawl-walk-run starting with PINRemoteImage's cache-control implementation. I think the benefits of cached decompressed images probably outweighs the downsides of _over_-requesting of remote images.

PR #479, opened 2018-09-21
Proper check to only call +defaultImageCache if it's overridden

nguyenhuy:
The current check (`-respondsToSelector:`) always returns `YES` because it's implemented [here](https://github.com/pinterest/PINRemoteImage/blob/master/Source/Classes/PINRemoteImageManager.m#L239).

### DIFF
--- a/Source/Classes/PINRemoteImageManager.m
+++ b/Source/Classes/PINRemoteImageManager.m
@@ -11,6 +11,8 @@
 #import <CommonCrypto/CommonDigest.h>
 #import <PINOperation/PINOperation.h>
 
+#import <objc/runtime.h>
+
 #import "PINAlternateRepresentationProvider.h"
 #import "PINRemoteImage.h"
 #import "PINRemoteLock.h"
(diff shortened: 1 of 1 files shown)


pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

nguyenhuy:
cc @wiseoldduck

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/65f21d263d4d48adfe8eef02b5f80338/raw/044ae19f41fc0617a5d1c45c5c99b08859d4df92/log.txt)

Source/Classes/PINRemoteImageManager.m
```diff
@@ -79,6 +81,16 @@ float dataTaskPriorityWithImageManagerPriority(PINRemoteImageManagerPriority pri
     }
 }
 
+// Reference: https://github.com/TextureGroup/Texture/blob/5dd5611/Source/Private/ASInternalHelpers.m#L60
+BOOL PINRemoteImageManagerSubclassOverridesSelector(Class subclass, SEL selector)
```
maicki: We may should move this to a central place if there is one already.

nguyenhuy: There is none that I'm aware of. Let's do it next time we have a similar need.

Source/Classes/PINRemoteImageManager.m
```diff
@@ -171,13 +183,13 @@ - (nonnull instancetype)initWithSessionConfiguration:(nullable NSURLSessionConfi
     if (self = [super init]) {
         if (imageCache) {
             self.cache = imageCache;
+        } else if (PINRemoteImageManagerSubclassOverridesSelector([self class], @selector(defaultImageCache))) {
```
wiseoldduck: 🙈 oh no! Thanks for the fix...

nguyenhuy: No problem at all!

Source/Classes/PINRemoteImageManager.m
```diff
@@ -233,6 +245,7 @@ - (nonnull instancetype)initWithSessionConfiguration:(nullable NSURLSessionConfi
 }
 
 - (id<PINRemoteImageCaching>)defaultImageCache {
+    NSAssert(NO, @"Assert should not have been reached");
```
wiseoldduck: Actually, I don't think this can be asserted because I think it would be legitimate for a subclass to override the (deprecated) instance method and call `[super default...]` with the intent of customizing it.

nguyenhuy: Good point. Updated!

PR #500, opened 2019-03-11
add cache options when downloading images

ernestmama:
Adding cache options when download images or pre-warm media connections. When we are prefetching, we do not necessarily want to save in the disk or even save it at all (also there is an issue in the disk cache which cause a crash when running in processEntry which is called in the +main).

### DIFF
--- a/Source/Classes/PINRemoteImageManager.h
+++ b/Source/Classes/PINRemoteImageManager.h
@@ -45,6 +45,17 @@ typedef NS_ERROR_ENUM(PINRemoteImageManagerErrorDomain, PINRemoteImageManagerErr
 };
 
 /**
+ Options with which to cache image in memory / disk / both
+ */
+typedef NS_OPTIONS(NSUInteger, PINRemoteImageManagerCacheOptions) {
+    PINRemoteImageManagerCacheOptionsNone = 0,
+    /** Store images only in disk */
+    PINRemoteImageManagerCacheOptionsDisk = 1,
+    /** Store images only in memory */
+    PINRemoteImageManagerCacheOptionsMemory = 1 << 1
+};
+
+/**
  Options with which to download and process images
  */
 typedef NS_OPTIONS(NSUInteger, PINRemoteImageManagerDownloadOptions) {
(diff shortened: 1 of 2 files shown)


Source/Classes/PINRemoteImageManager.h
```diff
@@ -44,6 +44,17 @@ typedef NS_ERROR_ENUM(PINRemoteImageManagerErrorDomain, PINRemoteImageManagerErr
     PINRemoteImageManagerErrorImageEmpty = 5,
 };
 
+/**
+ Options with which to cache image in memory / disk / both
```
rqueue: nit: should `both` be `none`?

Source/Classes/PINRemoteImageManager.m
```diff
@@ -1361,6 +1502,40 @@ - (BOOL)materializeAndCacheObject:(id)object
...
+                        outAltRep:(id *)outAlternateRepresentation
+{
+    return [self materializeAndCacheObject:object
+                               cacheInDisk:diskData
+                            additionalCost:additionalCost
+                                    maxAge:nil
+                                       url:url
+                                       key:key
+                                   options:options
+                              cacheOptions:PINRemoteImageManagerCacheOptionsMemory | PINRemoteImageManagerCacheOptionsDisk
```
rqueue: What do you think about adding a `PINRemoteImageManagerCacheOptionsMemory` default option, which can be set to be `PINRemoteImageManagerCacheOptionsMemory | PINRemoteImageManagerCacheOptionsDisk`? I think it would make it easier in the future if we needed to make method calls that use the default behavior or if we ever wanted to change the default behavior altogether

maicki:
What do you all think about finally adding an object as parameter. I don't think adding another parameter is a good idea. We could just add one new initializer and a new option class that calls the old one by extracting the options arguments from it. cc @garrettmoon

Source/Classes/PINRemoteImageManager.m
ernestmama: Changing the default behavior would affect the existing users of the framework. Thats why I am keeping both as the default.

rqueue: Oh yea I wasn’t suggesting to change the default behavior. Sorry, just realized a typo in my comment. I meant to suggest a PINRemoteImageManagerCacheOptionDefault Option, whoa value would be PINRemoteImageManagerCacheOptionMemory | PINRemoteImageManagerCacheOptionDisk

rqueue:
@maicki I'm a fan of that idea. It would make it easier to add/remove parameters in the future as well as make it easier to see all available parameters from a method call via seeing the config/option object class (assuming in the future we move towards only using the object as a parameter)

pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/ec3408496f18f3e801c00b699f369f86/raw/673cb0dc87184318ae450b9ec76b016922a1425b/log.txt)

sa-mu:
Sa-mu muito bom

ernestmama:
will put up another pr later with latest updates. close for now.

PR #501, opened 2019-05-02
Fix webp animated image frame over lapping

wsdwsd0829:
The webp interpreter is trying to draw both previous and current frame to context resulting:
Image_11
https://lh3.googleusercontent.com/z2B73tA2ofYnFho0sj2KEQpze1fYY7RFSh5Wa9uElu67vF0R6wB59S2Roe3mzqbca5VHA2EaTTmEjtIxyeN9=s208-rwa

### DIFF
--- a/Source/Classes/AnimatedImages/PINWebPAnimatedImage.m
+++ b/Source/Classes/AnimatedImages/PINWebPAnimatedImage.m
@@ -171,7 +171,7 @@ static void releaseData(void *info, const void *data, size_t size)
     }
     
     if (data) {
-        CGDataProviderRef provider = CGDataProviderCreateWithData(NULL, data, _width * _height * pixelLength, releaseData);
+        CGDataProviderRef provider = CGDataProviderCreateWithData(NULL, data, iterator.width * iterator.height * pixelLength, releaseData);
         
         CGColorSpaceRef colorSpaceRef = CGColorSpaceCreateDeviceRGB();
         CGBitmapInfo bitmapInfo = kCGBitmapByteOrderDefault;
--- a/Source/Classes/AnimatedImages/PINWebPAnimatedImage.m
+++ b/Source/Classes/AnimatedImages/PINWebPAnimatedImage.m
@@ -14,6 +14,71 @@
 
 #import "webp/demux.h"
 
+@interface PINAnimatedWebPFrameInfo : NSObject <NSCopying>
+
+/**
+ * Each frame of an animated WebP may cover only a portion of the full image.
+ * `frameRect` records what portion of the image this frame covers
+ */
+@property (nonatomic, readonly) CGRect frameRect;
+
+/**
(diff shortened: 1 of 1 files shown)


pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

wsdwsd0829:
This pr just to start the conversation. 
I feel like we do not need to do the extra work? 
https://github.com/pinterest/PINRemoteImage/blob/master/Source/Classes/AnimatedImages/PINWebPAnimatedImage.m#L142 but there may be other cases depends on that? 
So 
option1: I suggest remove the previousFrame parameter if it's safe to do so.  (preferred)
option2: add a bool to control it so we can disable it, where do u suggest to control it best? (maybe ASAnimatedImageProtocol in ASImageProtocols.h ?)

garrettmoon:
Hmm, my understanding of the spec is you can only not draw the previous frame if the current frame is a key frame. And that is taken into account in the code currently. I wonder if there's some other issue with transparency or background color?

wsdwsd0829:
I tested a couple of images and seems fine, do you have any examples for me to test against. 
the issue is we cannot draw previous and current frame together. (https://github.com/pinterest/PINRemoteImage/blob/master/Source/Classes/AnimatedImages/PINWebPAnimatedImage.m#L142)  which will result in the image in description.

wsdwsd0829:
Earlier attempt do break some cases because image blending logic is much more complex than expected. 
frameRequiresBlendingWithPreviousFrame is borrowed from some other framework.
I tested about 20 images that either/both breaks with current lib or first attempt. 
After the fix all looks good, I will post the test links once get permission :)

pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

garrettmoon:
@wsdwsd0829 Lemme know when you get a chance to post the test images!

wsdwsd0829:
Sent to your email.

Source/Classes/AnimatedImages/PINWebPAnimatedImage.m
```diff
@@ -278,6 +385,38 @@ - (CGImageRef)imageAtIndex:(NSUInteger)index cacheProvider:(nullable id<PINCache
     return canvas;
 }
 
+- (BOOL)frameRequiresBlendingWithPreviousFrame:(NSUInteger)index
```
garrettmoon: This seems very similar to the isKeyFrame logic… can that just be updated instead?

garrettmoon:
I'm going to close this for now, lemme know if I missed anything in my PR!

PR #514, opened 2019-07-08
Fixes re-downloading data corrupt for the same url

### LINKED ISSUE #489: The data received does not correctly match the corresponding PINRemoteImageDownloadTask
Image data is received piece by piece while the image is being downloaded. In the process of image data download, I quickly back and forth scroll the view, the current download task will be cancelled, PINRemoteImageDownloadTask will be removed from the tasks, and then start a new task. This looks fine,  but the problem occurs when the URL of the new task is the same as the URL of the cancelled task.
Due to image data is returned piece by piece, it is possible that the server returning a piece of data. At this time, the task of this URL has been cancelled, and the PINRemoteImageDownloadTask of corresponding URL(as key) in tasks will be removed. And then start the download task of same URL, the tasks will store the PINRemoteImageDownloadTask of corresponding URL(as key) again. Now, the data of cancelled task was received, PINURLSessionManagerDelegate method according to PINRemoteImageCacheKey find the cacheKey(URL) in NSURLProtocol, and according to cacheKey find the PINRemoteImageDownloadTask in the tasks. This PINRemoteImageDownloadTask should belong to the new task, and the received data should belong to the cancelled task, the received data data will be added to the PINRemoteImageDownloadTask. This will be stored with the correct data for the new task, resulting in an image decoding exception.

```objc
// PINRemoteImageManager.m
- (void)didReceiveData:(NSData *)data forTask:(NSURLSessionDataTask *)dataTask
{
    [self lock];
        NSString *cacheKey = [NSURLProtocol propertyForKey:PINRemoteImageCacheKey inRequest:dataTask.originalRequest];
        PINRemoteImageDownloadTask *task = [self.tasks objectForKey:cacheKey];
    
        /********************************** Here is the NSAssert I added ***************************************************/
        NSAssert(dataTask.taskIdentifier == task.urlSessionTaskOperation.dataTask.taskIdentifier, @"The received data belong to cancelled task.")
        /*******************************************************************************************************************/
    
        if (task.progressImage == nil) {
            task.progressImage = [[PINProgressiveImage alloc] init];
            task.progressImage.startTime = task.sessionTaskStartTime;
            task.progressImage.estimatedRemainingTimeThreshold = self.estimatedRemainingTimeThreshold;
            if (self.progressThresholds) {
            }
        }

        PINProgressiveImage *progressiveImage = task.progressImage;
        BOOL hasProgressBlocks = task.hasProgressBlocks;
        BOOL shouldBlur = self.shouldBlurProgressive;
        CGSize maxProgressiveRenderSize = self.maxProgressiveRenderSize;
    
        [task callProgressDownloadWithQueue:self.callbackQueue completedBytes:dataTask.countOfBytesReceived totalBytes:dataTask.countOfBytesExpectedToReceive];
    [self unlock];
    
    [progressiveImage updateProgressiveImageWithData:data expectedNumberOfBytes:[dataTask countOfBytesExpectedToReceive]];

    if (hasProgressBlocks && [NSOperation instancesRespondToSelector:@selector(qualityOfService)]) {
        __weak typeof(self) weakSelf = self;
        [_concurrentOperationQueue pin_addOperationWithQueuePriority:PINRemoteImageManagerPriorityLow block:^{
            typeof(self) strongSelf = weakSelf;
            CGFloat renderedImageQuality = 1.0;
            PINImage *progressImage = [progressiveImage currentImageBlurred:shouldBlur maxProgressiveRenderSize:maxProgressiveRenderSize renderedImageQuality:&renderedImageQuality];
            if (progressImage) {
                [strongSelf lock];
                    PINRemoteImageDownloadTask *task = strongSelf.tasks[cacheKey];
                    [task callProgressImageWithQueue:strongSelf.callbackQueue withImage:progressImage renderedImageQuality:renderedImageQuality];
                [strongSelf unlock];
            }
        }];
    }
}
```

I added an assertion, and then when I scroll back and forth quickly the view has a probability that it will hit.

zhongwuzw:
Fixes #489 , we use `cacheKey` to get the `task`, actually it would mix up `data` if previous task cancelled but download delegate still not finished , and next task with the same url started at the same time. This is the race condition for the same url downloading, we can keep the weak reference to `task` and get it in session delegate.


pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

bolsinga:
Please add a test.

garrettmoon:
This is a really great fix! Would you mind adding a test as @bolsinga suggested?

zhongwuzw:
@bolsinga @garrettmoon Please review again. :)

pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

bolsinga:
Thanks for the test!

Tests/PINRemoteImageTests.m
```diff
@@ -204,6 +234,18 @@ - (NSURL *)imageFrom404URL
 - (void)didReceiveData:(NSData *)data forTask:(NSURLSessionTask *)task
 {
     self.task = task;
+    if (self.semaphore) {
```
garrettmoon: Can you add a comment saying which test this is for?

Tests/PINRemoteImageTests.m
```diff
@@ -81,8 +109,10 @@ @interface PINRemoteImage_Tests : XCTestCase <PINURLSessionManagerDelegate>
 @property (nonatomic, strong) PINRemoteImageManager *imageManager;
 @property (nonatomic, strong) NSMutableData *data;
 @property (nonatomic, strong) NSURLSessionTask *task;
+@property (nonatomic, strong) PINRemoteImageDownloadTask *firstDownloadTask;
+@property (nonatomic, strong) PINRemoteImageDownloadTask *secondDownloadTask;
 @property (nonatomic, strong) NSError *error;
-
+@property (nonatomic, strong) dispatch_semaphore_t semaphore;
```
garrettmoon: Should the new properties be nil'd out in tearDown?

garrettmoon:
Can you add an entry to CHANGELOG.md for this?

Tests/PINRemoteImageTests.m
zhongwuzw: 🤔 We should.

Tests/PINRemoteImageTests.m
zhongwuzw: Sure. 👍

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/14f6d7a1a8bcabcbdbf6c1f1deb20a7d/raw/1adbfbc32cc36ca800ad932a676cb02075a93b47/log.txt)

garrettmoon:
Mind rebasing against master? I think I fixed a flakey test.

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/37df558737d307c1dacce9437db4adc0/raw/5e7ef88a7e314cb0440e59ea6ee5eb11bb6c9ced/log.txt)

zhongwuzw:
@garrettmoon Tests failed like below frequently, I think we may need to add a command like `rm -rf ~/Library/Developer/Xcode/DerivedData/` before build project in CI?

image

jparise:
Sorry for the long delay @zhongwuzw. The CI tests should now be fixed. If you could rebase again, we can do a final review of your change.

zhongwuzw:
@jparise Done.

Source/Classes/PINRemoteImageManager.m
```diff
@@ -97,6 +97,32 @@ BOOL PINRemoteImageManagerSubclassOverridesSelector(Class subclass, SEL selector
 NSString * const PINRemoteImageCacheKeyResumePrefix = @"R-";
 typedef void (^PINRemoteImageManagerDataCompletion)(NSData *data, NSURLResponse *response, NSError *error);
 
+@interface PINRemoteImageWeakContainer : NSObject
+
+@property (nonatomic, readonly, weak) id target;
+
++ (PINRemoteImageWeakContainer *)weakContainerWithTarget:(id)target;
```
jparise: Because this is only used internally, we could save the extra code + symbols by removing this class method and just use `-initWithTarget:` in the one place we call it.

Source/Classes/PINRemoteImageManager.m
```diff
@@ -97,6 +97,32 @@ BOOL PINRemoteImageManagerSubclassOverridesSelector(Class subclass, SEL selector
...
+@implementation PINRemoteImageWeakContainer
+
++ (PINRemoteImageWeakContainer *)weakContainerWithTarget:(id)target
+{
+    return [[self alloc] initWithTarget:target];
+}
+
+- (instancetype)initWithTarget:(id)target
+{
+    if (self) {
```
jparise: `[super init]`?

Tests/PINRemoteImageTests.m
```diff
@@ -60,6 +61,33 @@ - (void)swizzled_scheduleDownloadWithRequest:(NSURLRequest *)request
...
+@implementation PINRemoteImageWeakContainer
+
++ (PINRemoteImageWeakContainer *)weakContainerWithTarget:(id)target
+{
+    return [[self alloc] initWithTarget:target];
+}
+
+- (instancetype)initWithTarget:(id)target
+{
+    if (self) {
```
jparise: `[super init]`?

bolsinga: I'd also not expect to re-define this class in the tests...

bolsinga: Why is this class duplicated in the test? I don't think that should happen,

Source/Classes/PINRemoteImageManager.m
```diff
@@ -97,6 +97,32 @@ BOOL PINRemoteImageManagerSubclassOverridesSelector(Class subclass, SEL selector
...
+@interface PINRemoteImageWeakContainer : NSObject
+
+@property (nonatomic, readonly, weak) id target;
+
++ (PINRemoteImageWeakContainer *)weakContainerWithTarget:(id)target;
+- (instancetype)initWithTarget:(id)target;
+
+@end
+
+@implementation PINRemoteImageWeakContainer
```
bolsinga: I think this can have a better name.

Source/Classes/PINRemoteImageManager.m
```diff
@@ -97,6 +97,32 @@ BOOL PINRemoteImageManagerSubclassOverridesSelector(Class subclass, SEL selector
...
+
++ (PINRemoteImageWeakContainer *)weakContainerWithTarget:(id)target
```
bolsinga: Instead of using `id`, please use a real type so the compiler can help us out if we get something wrong.

Source/Classes/PINRemoteImageManager.m
```diff
@@ -97,6 +97,32 @@ BOOL PINRemoteImageManagerSubclassOverridesSelector(Class subclass, SEL selector
 NSString * const PINRemoteImageCacheKeyResumePrefix = @"R-";
 typedef void (^PINRemoteImageManagerDataCompletion)(NSData *data, NSURLResponse *response, NSError *error);
 
+@interface PINRemoteImageWeakContainer : NSObject
+
+@property (nonatomic, readonly, weak) id target;
```
bolsinga: Please use a type for this property

Tests/PINRemoteImageTests.m
```diff
@@ -1418,6 +1468,28 @@ - (void)testRetry
...
+    firstTask = self.firstDownloadTask;
+
+    [self.imageManager cancelTaskWithUUID:firstUUID];
+    
+    PINRemoteImageTask *secondTask = nil;
+    [self.imageManager downloadImageWithURL:[self transparentPNGURL] options:0 progressDownload:nil completion:nil];
+    dispatch_semaphore_wait(self.semaphore, [self timeout]);
+    secondTask = self.secondDownloadTask;
+    
+    XCTAssert(firstTask != secondTask);
```
bolsinga: So I think this test will work but it involves caching the task, and not the key/request. So this test would not compile let alone fail w/o your change. Is it possible to add a test that will compile w/o the fix and fail to pass, and compile w/ the fix and pass the test? Thanks.

Source/Classes/PINRemoteImageManager.m
zhongwuzw: Good catch 👍 , no harm but I'll fix it.

Tests/PINRemoteImageTests.m
zhongwuzw: Emm, sorry my English is poor, I may don't understand it.

Source/Classes/PINRemoteImageManager.m
jparise: We could name this `PINRemoteImageWeakTask` now that it always holds `PINRemoteImageTask` objects.

Source/Classes/PINRemoteImageManager.m
```diff
@@ -97,6 +97,26 @@ BOOL PINRemoteImageManagerSubclassOverridesSelector(Class subclass, SEL selector
 NSString * const PINRemoteImageCacheKeyResumePrefix = @"R-";
 typedef void (^PINRemoteImageManagerDataCompletion)(NSData *data, NSURLResponse *response, NSError *error);
 
+@interface PINRemoteImageWeakContainer : NSObject
+
+@property (nonatomic, readonly, weak) PINRemoteImageTask *target;
+
+- (instancetype)initWithTarget:(PINRemoteImageTask *)target;
```
jparise:
```suggestion
- (instancetype)initWithTask:(PINRemoteImageTask *)task;
```

Source/Classes/PINRemoteImageManager.m
```diff
@@ -97,6 +97,26 @@ BOOL PINRemoteImageManagerSubclassOverridesSelector(Class subclass, SEL selector
 NSString * const PINRemoteImageCacheKeyResumePrefix = @"R-";
 typedef void (^PINRemoteImageManagerDataCompletion)(NSData *data, NSURLResponse *response, NSError *error);
 
+@interface PINRemoteImageWeakContainer : NSObject
+
+@property (nonatomic, readonly, weak) PINRemoteImageTask *target;
```
jparise:
```suggestion
@property (nonatomic, readonly, weak) PINRemoteImageTask *task;
```

Source/Classes/PINRemoteImageManager.m
```diff
@@ -955,7 +975,8 @@ - (NSURLRequest *)requestWithURL:(NSURL *)url key:(NSString *)key
         request = [_requestConfigurationHandler(request) mutableCopy];
     }
     
-    [NSURLProtocol setProperty:key forKey:PINRemoteImageCacheKey inRequest:request];
+    PINRemoteImageWeakTask *container = [[PINRemoteImageWeakTask alloc] initWithTask:task];
+    [NSURLProtocol setProperty:container forKey:PINRemoteImageCacheKey inRequest:request];
```
jparise: I think we should also update these calls because we're no longer storing the key as a property. We now store a reference to a `PINRemoteImageWeakTask *`.

So let's rename `PINRemoteImageCacheKey` to `PINRemoteImageWeakTaskKey` (and set its value to `@"PINRemoteImageWeakTaskKey"`. Also, let's rename the `container` variables to `weakTask` for complete consistency.

zhongwuzw:
> This is looking really good, @zhongwuzw! Just a few more bits of feedback.

Done.

Tests/PINRemoteImageTests.m
garrettmoon: @bolsinga I think testing this will be exceedingly difficult without a testing network stack we can control. For now, let's merge this and think about how we can create a mocked networking stack since it would make *all* of our tests much easier to write.

garrettmoon:
This looks great, thank you so much for the fix!

Tests/PINRemoteImageTests.m
bolsinga: ok

PR #528, opened 2019-08-02
Removes the optimization to reduce the display links frame interval

garrettmoon:
The idea behind this optimization was to reduce the number of calls the display
link would need for animated images which didn't draw every frame. But this
has resulted in frames being dropped in certain cases where frameInterval isn't
respected (see https://i.pinimg.com/originals/58/9c/e8/589ce828d2ef0797da0e85b1c6d2f2b3.gif)

Since the the display link fire method continues if we detect we're
at the same image index as before, let's just remove this optimization.

### DIFF
--- a/Source/Classes/AnimatedImages/PINAnimatedImageView.m
+++ b/Source/Classes/AnimatedImages/PINAnimatedImageView.m
@@ -211,14 +211,10 @@
     if ([self canBeVisible] == NO) {
         return;
     }
-    
-    // Get frame interval before holding display link lock to avoid deadlock
-    NSUInteger frameInterval = self.animatedImage.frameInterval;
 
     if (_displayLink == nil) {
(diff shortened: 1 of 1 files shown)


pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/d8e370c40b34b899567f9aee30c19e1c/raw/39620f5a6a95ddf533fd48e9f58042f7c478fa17/log.txt)

bolsinga:
Is it possible to add a test covering this change?

PR #540, opened 2020-02-23
#trivial Fix: Download fail when using processor with GIF image

Naituw:
Before: Since GIF image become alternate representation, processor can't get an UIImage/NSImage to process, a `PINRemoteImageManagerErrorFailedToFetchImageForProcessing` error is returned.

This Fix: Disallow alternate representation when obtaining the image for processing.

### DIFF
--- a/Source/Classes/PINRemoteImageManager.m
+++ b/Source/Classes/PINRemoteImageManager.m
@@ -776,7 +776,7 @@ static dispatch_once_t sharedDispatchToken;
         
         __weak typeof(self) weakSelf = self;
         NSUUID *downloadTaskUUID = [self downloadImageWithURL:url
-                                                      options:options | PINRemoteImageManagerDownloadOptionsSkipEarlyCheck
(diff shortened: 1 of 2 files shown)


pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/03168c0b3bd93db0c6c4b6d17a56875e/raw/09b475ccdfa3d7bfee6a1b27a156b386959aef9e/log.txt)

strangeliu:
For gifs, alternate representation is enabled by default, Use `downloadImageWithURL:options:completion:` to customize options.

Naituw:
> For gifs, alternate representation is enabled by default, Use `downloadImageWithURL:options:completion:` to customize options.

Current implementation will fail on the test I added.

I don't mean to disable alternate representation by default for gif, but alternative reps should not created when user want to process the image by a processor.

garrettmoon:
Can you please rebase this? (sorry for the delayed review!)

PR #549, opened 2020-05-07
Integrated UIGraphicsImageRenderer for iOS/tvOS 10.0 and above

IHEARTCOOKIES:
Migrated to `UIGraphicsImageRenderer` for applications running iOS/tvOS 10.0 and above. 

As mentioned in this [WWDC talk](https://developer.apple.com/videos/play/wwdc2018/416/?time=1269), `UIGraphicsImageRenderer` allows us to punt the BPP (bits per pixel), et. al, to the system. As a result, this allows us to save up to [75% less memory](https://developer.apple.com/videos/play/wwdc2018-416/?time=1230) in cases we previously wouldn't.

The implementation defaults back to the original way of decoding if the user is on an older version of iOS/tvOS (below 10.0).

### DIFF
--- a/Source/Classes/Categories/PINImage+DecodedImage.m
+++ b/Source/Classes/Categories/PINImage+DecodedImage.m
@@ -16,6 +16,49 @@
 
 #import "NSData+ImageDetectors.h"
 
+NS_INLINE BOOL pin_CGImageRefIsOpaque(CGImageRef imageRef) {
+    CGImageAlphaInfo alpha = CGImageGetAlphaInfo(imageRef);
+    switch (alpha) {
+        case kCGImageAlphaNone:
+        case kCGImageAlphaNoneSkipLast:
+        case kCGImageAlphaNoneSkipFirst:
+            return YES;
+        default:
+            return NO;
+    }
+}
+
+#if PIN_TARGET_IOS
+NS_INLINE void pin_degreesFromOrientation(UIImageOrientation orientation, void (^completion)(CGFloat degrees, BOOL horizontalFlip, BOOL verticalFlip)) {
+    switch (orientation) {
+        case UIImageOrientationUp: // default orientation
+            completion(0.0, NO, NO);
+            break;
+        case UIImageOrientationDown: // 180 deg rotation
+            completion(180.0, NO, NO);
+            break;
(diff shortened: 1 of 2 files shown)


IHEARTCOOKIES:
Sorry for the delay @bolsinga.

I added two tests that exercise the `UIGraphicsImageRenderer` behavior 👍 

Let me know what you think

Source/Classes/Categories/PINImage+DecodedImage.m
```diff
@@ -224,6 +277,47 @@ UIImageOrientation pin_UIImageOrientationFromImageSource(CGImageSourceRef imageS
     return orientation;
 }
 
+void pin_degreesFromOrientation(UIImageOrientation orientation, void (^completion)(CGFloat degrees, BOOL horizontalFlip, BOOL verticalFlip)) {
```
bolsinga, line 280: Please use `NS_INLINE` and move it to before it is used near the top of the file.

Source/Classes/Categories/PINImage+DecodedImage.m
```diff
@@ -224,6 +277,47 @@ UIImageOrientation pin_UIImageOrientationFromImageSource(CGImageSourceRef imageS
...
+            break;
+        case UIImageOrientationRightMirrored: // vertical flip
+            completion(90.0, NO, YES);
+            break;
+    }
+}
+
 #endif
 
+BOOL pin_CGImageRefIsOpaque(CGImageRef imageRef) {
```
bolsinga, line 311: Please `NS_INLINE` and move towards the top of the file.

bolsinga (approved):
Thanks for the test! I approve if you please `NS_INLINE` those helper methods.

Source/Classes/Categories/PINImage+DecodedImage.m
IHEARTCOOKIES: Certainly!

Source/Classes/Categories/PINImage+DecodedImage.m
IHEARTCOOKIES: Consider it done 👍

bolsinga (approved):

PR #553, opened 2020-06-24
Remove the PCH file and ensure macros are defined before using

bolsinga:


bolsinga:
non-fast-forward nonsense

PR #556, opened 2020-07-02
Remove the PCH file and ensure macros are defined before using

bolsinga:


bolsinga:
> Does the Availability import need to go into macros?

Good Q. I just had it import <Foundation/Foundation.h> to get it all settled.

PR #558, opened 2020-08-04
Fix compiling in Xcode 12.0b4

zacwest:
This beta added a diagnostic for passing more specific types to `id`-typed values:

> Fixed type checking for block parameters using id with protocols. The compiler now emits an error for method calls with a block that uses parameters more specific than arguments with which it will be called. (57980961)


mycroftcanner:
Thanks @zacwest 

 @garrettmoon @mRs- @Adlai-Holler

wsmurphy:
Should this PR also update the version of PINCache to the version with the same fix for 12b4?

PR #560, opened 2020-08-07
This warning shows up downstream

garrettmoon:


bolsinga:
I'd tried getting this in #552 and #556. I hope we can fix it instead of ignore it!

garrettmoon:
@bolsinga cool, I'll let you have this one, thanks for figuring it out!

Issue #559, opened 2020-08-06
Xcode 12 beta 4 can't compile you


mr-fixit:
FYI,

Xcode seems to have trouble figuring out that various classes that support the PINCaching protocol should be allowed as parameters of type `id<PINCaching>`.

mycroftcanner:
Can't compile either...

```
Incompatible block pointer types sending 'void (^)(PINCache * _Nonnull __strong, NSString * _Nonnull __strong, id  _Nullable __strong)' to parameter of type 'PINCacheObjectBlock _Nullable' (aka 'void (^)(id<PINCaching>  _Nonnull __strong, NSString * _Nonnull __strong, id  _Nullable __strong)')
```

mycroftcanner:
@zacwest @garrettmoon @mRs- @Adlai-Holler Any chance you could commit a fix for PINRemoteImage please?

https://github.com/pinterest/PINCache/pull/273

zboralski:
### Temporary Workaround 
```
ENV['COCOAPODS_DISABLE_STATS'] = "true"
use_frameworks!

target 'XX' do
    platform :ios, '13.6'
    pod 'Texture', :git => 'https://github.com/TextureGroup/Texture', :branch => 'master'
    pod 'PINRemoteImage', :git => 'https://github.com/zacwest/PINRemoteImage', :branch => 'fix-xcode-12.0b4'
    pod 'PINCache', :git => 'https://github.com/pinterest/PINCache', :branch => 'master'
end
```

Destriarch:
Still getting this issue in PINRemoteImage 3.0.3

PR #562, opened 2020-08-21
Update PINCache and PINOperation

garrettmoon:
Fixes https://github.com/pinterest/PINRemoteImage/issues/559


mr-fixit:
What's the ETA for publishing 3.0.1?

PR #571, opened 2020-10-01
SPM support

3a4oT:
This is POC.

- Depends on https://github.com/pinterest/PINCache/pull/283.

- Use only for SPM https://github.com/SDWebImage/libwebp-Xcode - `libwebp` wrapper. (submodule pointed to libwebp GitHub mirror)

- The current `PINRemoteImage` source files layout is hard to adopt for SPM. To allow build with SPM I added the `include` folder and moved all public headers to it.

- SPM test target is omitted since currently, PINRemoteImageTests contains mixed sources (Objective-C and Swift) which are not supported by SPM. Tests sources should be restructured in the way we could make two separate test targets for Swift sources and Objective-C. As far as we can perform testing using the Xcode project to run tests I think it's fine to leave it as is.

 - Carthage doesn't support Xcode 12 yet. Pinned to Xcode 11.7


garrettmoon:
@3a4oT Looks good to me! Thank you for putting up this PR!

garrettmoon (approved):

3a4oT:
@garrettmoon thanks! Can you please review follow up commits  #573?

PR #572, opened 2020-10-06
Use proper units when setting preferredFramesPerSecond

### LINKED ISSUE #570: GIF playback is slow/broken
GIFs are render extremely slowly in `3.0.1`. 

Here is the example GIF in the Swift Example app provided in the repo versus the same GIF loaded in Chrome: 

slow_gif

This can be replicated with other GIFs also. 

Setup: 
- PINRemoteImage `3.0.1`
- Xcode 12 
- iPhone 11 Pro, iOS 14 simulator 

Any assistance would be greatly appreciated 🙏  Thanks.

bdolman:
preferredFramesPerSecond expects frames-per-second, rather than the
fractional interval expected by frameInterval.

For example, frameInterval=2 is the equivalent (on most devices) of
preferredFramesPerSecond=30.

This issue was introduced in https://github.com/pinterest/PINRemoteImage/pull/552

It's likely this fixes #570

### DIFF
--- a/Source/Classes/AnimatedImages/PINAnimatedImageView.m
+++ b/Source/Classes/AnimatedImages/PINAnimatedImageView.m
@@ -221,13 +221,16 @@
     
     // Get frame interval before holding display link lock to avoid deadlock
     NSUInteger frameInterval = self.animatedImage.frameInterval;
+    
+    // Convert from display link fractional value to fps (note: frameInterval is always at least 1)
+    NSInteger frameRate = ceil([PINAnimatedImage maximumFramesPerSecond] / ((double) frameInterval));
 
     if (_displayLink == nil) {
         _playHead = 0;
         _displayLink = [PINDisplayLink displayLinkWithTarget:[PINRemoteWeakProxy weakProxyWithTarget:self] selector:@selector(displayLinkFired:)];
 #if PIN_TARGET_IOS
         if (@available(iOS 10.0, tvOS 10.0, *)) {
-            _displayLink.preferredFramesPerSecond = frameInterval;
+            _displayLink.preferredFramesPerSecond = frameRate;
         } else {
 #endif
             _displayLink.frameInterval = frameInterval;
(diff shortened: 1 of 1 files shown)


garrettmoon (approved):
Thank you for fixing!

garrettmoon:
Can you address:
```
    - WARN  | [PINRemoteImage/Core, PINRemoteImage/OSX, PINRemoteImage/WebP, and more...] xcodebuild:  /Users/runner/work/PINRemoteImage/PINRemoteImage/Source/Classes/AnimatedImages/PINAnimatedImageView.m:226:15: warning: unused variable 'frameRate' [-Wunused-variable]
29
```
And I'll merge?

bdolman:
@garrettmoon Sure thing. I would like to move the `frameRate` calculation closer to its use, like this:

```
#if PIN_TARGET_IOS
        if (@available(iOS 10.0, tvOS 10.0, *)) {
            // Convert from display link fractional value to fps (note: frameInterval is always at least 1)
            NSInteger frameRate = ceil([PINAnimatedImage maximumFramesPerSecond] / ((double) frameInterval));
            _displayLink.preferredFramesPerSecond = frameRate;
        } else {
#endif
```

Rather than doing the calculation up above when it might not be used, but I was concerned about this comment:

```
// Get frame interval before holding display link lock to avoid deadlock
    NSUInteger frameInterval = self.animatedImage.frameInterval;
```

I didn't know if I risked creating a deadlock by doing it after referencing the display lock. Do you know if that would be safe?

garrettmoon:
Ah! Please remove that comment, I believe it copy-pasta'd it from the version I wrote in Texture. It's not relevant here since this class is designed to access the displayLink only on main.

bdolman:
Perfect! I'll address that now.

bdolman:
I'm not sure why the `testResumeSkipCancelation` test failed, since it succeeded for me locally.

garrettmoon:
> I'm not sure why the `testResumeSkipCancelation` test failed, since it succeeded for me locally.

Seems to be flakey, I'll rerun.

PR #573, opened 2020-10-06
SPM - Objective-C++ consumers fix

3a4oT:
Workaround for Objective-C++ module which will consume this package, otherwise  - `Use of '@import' when C++ modules are disabled, consider using -fmodules and -fcxx-modules`. Texture is coming..


3a4oT:
@garrettmoon  I guess we need to change dependencies in `Package.swift`, currently it pointed to my fork instead if `Pinterest` repository. Will, you able to tag the new `PINCache` (3.0.2??) so I can point it to something like

`.package(url: "https://github.com/pinterest/PINCache.git", from: "3.0.2")`

garrettmoon:
Wow, what a terrible code reviewer I am! Yes, please update to Pinterest and I'll publish a new release.

garrettmoon:
> @garrettmoon I guess we need to change dependencies in `Package.swift`, currently it pointed to my fork instead of `Pinterest` repository. Will, you able to tag the new `PINCache` (3.0.2??) so I can point it to something like
> 
> `.package(url: "https://github.com/pinterest/PINCache.git", from: "3.0.2")`

3.0.2 is now available

3a4oT:
@garrettmoon Awesome! Should be ready for review! Thanks!

garrettmoon (approved):
Looks good to me! Thank you for putting this up!

PR #574, opened 2020-10-06
SPM define

3a4oT:
- allows building Texture as an Xcode SPM dependency 
 see https://github.com/TextureGroup/Texture/pull/1928
cc: @garrettmoon


garrettmoon (approved):
Looks good to me! Thanks for fixing!

PR #575, opened 2020-10-08
Moves to an .xcworkspace so we don't imbed subprojects

garrettmoon:
We may need to release PINCache and update the cartfile to point to it before we can merge this.


bolsinga (approved):

AYastrebov:
After this commit PINRemoteImage can't be install via carthage, because it has duplicate PINOperation in it

PR #579, opened 2020-10-19
Fix imports for non-SPM

garrettmoon:


rcancro (approved):
Changes look good to me!

PR #585, opened 2020-11-11
Update dependencies to fix carthage build

AYastrebov:


garrettmoon:
@AYastrebov thanks for putting this up! Does [this PR](https://github.com/pinterest/PINRemoteImage/pull/584) accomplish what you were trying to do? Is it fixed in master?

AYastrebov:
Yes, the problem was that I couldn't use `PINRemoteImage` installed with carthage, because of duplicate `PINOperation` dependency.

garrettmoon:
Thanks for confirming! I'll close this now.

PR #586, opened 2020-11-17
Fixed spm  integration on regular Xcode project

3a4oT:
- without this change project itself can be built as spm package, but failed to do so when you integrate it as a dependency via Xcode.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -56,3 +56,12 @@ jobs:
         uses: actions/checkout@v2
       - name: Verify that PINRemoteImage can be build by SPM
         run: make spm
+  example-spm-xcode:
+    runs-on: macos-latest
+    env:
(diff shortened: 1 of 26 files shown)


3a4oT:
Will appreciate a new release if possible :) There is a similar PR fo the https://github.com/pinterest/PINCache/pull/301
In general, spm/cocopods/xcode/carthage support is a little mind-blowing :)
cc @garrettmoon

garrettmoon:
Thanks for putting this up! I hate to ask, but would you mind updating the CI test to ensure that SPM works when integrating into a project?

3a4oT:
Yes sir! Will add it to my list.

garrettmoon (approved):
Thank you so much for putting this up, greatly appreciated!

mthole:
Can we get a release with this in it, please? @garrettmoon

Ektris:
Going to ask as well, can we please get a release for these fixes? @garrettmoon

PR #587, opened 2020-11-17
added Carthage to release pipeline

3a4oT:
- added Carthage step to release pipeline


garrettmoon:
Unfortunately I don't think this will work anymore since we need to build with --xcframeworks and Carthage doesn't support those: https://github.com/Carthage/Carthage/issues/3130

PR #588, opened 2020-11-19
Fix SPM support

woxtu:
Hi,

The current release installed by SPM can't import headers as follows:

```
fatal error: 'PINRemoteImage/PINRemoteImage.h' file not found
#import <PINRemoteImage/PINRemoteImage.h>
        ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

This PR contains the following changes to fix that:

- Put header files back next to source files
- Create new public headers directory


grangej:
This does seem to fix the build but it seems to break the categories.

woxtu:
@grangej
Thank you for your pointing out.
I could confirm that categories is not loaded, and maybe the current release has same issue.
Anyway I'll fix that.

grangej:
Awesome , any chance we can get this merged ??

mthole:
Bump?

mthole:
FYI: This may be taken care of by https://github.com/pinterest/PINRemoteImage/pull/586 already, which is merged to `master` but not in a release.

PR #592, opened 2020-12-07
Revert "Integrated UIGraphicsImageRenderer for iOS/tvOS 10.0"

IHEARTCOOKIES:
Passing large images (e.g, 5000x5000) to `UIGraphicsImageRenderer` seems to cause the underlying pipeline to allocate huge amounts of memory until the device finally OOMs on some arbitrary location, e.g:

```
PINRemoteImageManager, L1499:
*** Terminating app due to uncaught exception 'NSInternalInconsistencyException', reason: 'We only cache a decompressed image if we decompressed it ourselves. In that case, it should be backed by a CGImageRef.'
```

I've attached two screenshots of profiles (sorry, couldn't attach the actual profiles due to company policy) and created a radar for Apple to look into.

**Reproducing it should be relatively simple:**
1. Load large images, e.g, 5000x5000
2. Observe that the pipeline backing `UIGraphicsImageRenderer` grows in memory
3. Continue to load large images until the device finally OOMs

**Suggested action:**
Without spending too much time looking into the underlying reason (although I have some theories), I figured the most appropriate action would be to revert this integration until we better understand the bigger picture.

**Time profiler:**
Time profiler

**Allocations profiler:**
Allocations

### DIFF
--- a/Source/Classes/Categories/PINImage+DecodedImage.m
+++ b/Source/Classes/Categories/PINImage+DecodedImage.m
@@ -16,49 +16,6 @@
 
 #import "NSData+ImageDetectors.h"
 
-NS_INLINE BOOL pin_CGImageRefIsOpaque(CGImageRef imageRef) {
-    CGImageAlphaInfo alpha = CGImageGetAlphaInfo(imageRef);
-    switch (alpha) {
-        case kCGImageAlphaNone:
-        case kCGImageAlphaNoneSkipLast:
-        case kCGImageAlphaNoneSkipFirst:
-            return YES;
-        default:
-            return NO;
-    }
-}
-
-#if PIN_TARGET_IOS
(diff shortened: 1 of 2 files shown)


IHEARTCOOKIES:
cc: @bolsinga @garrettmoon

IHEARTCOOKIES:
FWIW, running the same test case without `UIGraphicsImageRenderer` still grows in memory - as expected (we're loading large images). But, the difference is in the watermarks. `UIGraphicsImageRenderer` seems to allocate a whole bunch of extra memory in its pipeline to best determine (bpp? format?), causing the process to spike up to abnormally high allocations.

I'm not completely sold on faulting `UIGraphicsImageRenderer`, however, lower/consistent watermarks is a win in my book. Let me know what you think @bolsinga @garrettmoon

PR #593, opened 2020-12-12
Fix typo in PINRemoteImageManager.html

HollowMan6:
pefetch -> prefetch


garrettmoon (approved):
Thanks for the improvement!

PR #594, opened 2021-01-06
Fix PINRemoteImageManager.h Close Comment Formatting

paulryanclark:
* The `*/` was not correctly placed on a new line. This prevents Xcode from correctly parsing the header file.


garrettmoon (approved):
LGTM! Thanks for the contribution!

PR #596, opened 2021-04-16
Fix Exif orientation unwinding

AlexQuinlivan:
Currently, the pin_degreesFromOrientation function that's used when decoding images returns the degrees that represent the Exif orientation. However, the rest of the decode function then applies those values to the already oriented image, causing it to turn the exact wrong way. This affected the normal and mirrored versions of Left and Right. 

The change in 6bbf499 addresses this by inverting the returned degrees to be used as a way to unwind the transform that's already applied to the current pixel grid. 

Hopefully this is a helpful PR for you, I've made some changes to the base example in 8fb0458 that allow for you to swap out the data source used. At 8fb0458 you can see the issue for the affected orientations, then by 6bbf499 they should all look correct.

### DIFF
--- a/Examples/Example-Shared/Kitten.h
+++ b/Examples/Example-Shared/Kitten.h
@@ -13,12 +13,8 @@
 #import <Cocoa/Cocoa.h>
 #endif
 
-@interface Kitten : NSObject
+#import "ImageSource.h"
 
-@property (nonatomic, strong) NSURL *imageURL;
-@property (nonatomic, strong) id dominantColor;
-@property (nonatomic, assign) CGSize imageSize;
-
-+ (void)fetchKittenForWidth:(CGFloat)width completion:(void (^)(NSArray *kittens))completion;
+@interface Kitten : NSObject <ImageSource>
 
 @end
--- a/Examples/Example-Shared/Kitten.m
+++ b/Examples/Example-Shared/Kitten.m
@@ -31,10 +31,12 @@
 
 #endif
 
-
 @implementation Kitten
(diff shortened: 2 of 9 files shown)


garrettmoon:
Would you mind adding a test which fails and is fixed by this PR?

AlexQuinlivan:
@garrettmoon sure thing, will get around to shortly

AlexQuinlivan:
@garrettmoon hopefully this test in 2b1b06a is sufficient. It creates a pixel grid and applies certain pixel layouts. It then runs it through the decode function to receive an image with up orientation. Fails on the four horizontal images when you revert 8fb0458, so that's good I guess. 

I also found another interesting bug along the way. UIGraphicsImageRenderer would handle CGFloats like `2.0000000004` by rounding up its output pixels to 3. So there's an extra commit in here that feeds it the correctly flipped image size, sans transform application. (Was picked up in the comparison of PNG data, took me a while to see the extra size...)

AlexQuinlivan:
@garrettmoon Hey mate, have you had a chance to look over any of this?

JunyuKuang:
Thank you, this works like a charm! I wonder why it hasn't got merged yet.

AlexQuinlivan:
@JunyuKuang no idea. But I've been running this in production now for two years with success.

mikengyn:
This fixed an issue I was dealing with. Thank you!

esme-putt:
Running this fix in production, and it's working great 👌 would love to see it merged!

PR #599, opened 2021-05-26
Remove carthage folder

AYastrebov:
- Remove carthage folder from repo (the same as https://github.com/pinterest/PINCache/pull/306)
- Use https://github.com/SDWebImage/libwebp-Xcode as webp dependency


Cartfile
```diff
@@ -1,3 +1,3 @@
-git "https://chromium.googlesource.com/webm/libwebp" "v1.1.0"
```
garrettmoon, line 1: Can you put this in a separate PR?

garrettmoon:
Like the PINCache PR I don't think we can just remove Carthage – the current setup has the project built with it.

Cartfile
AYastrebov: Sure, will make it tomorrow

AYastrebov:
> Like the PINCache PR I don't think we can just remove Carthage – the current setup has the project built with it.

Ok, I'll take a closer look at that.

AYastrebov:
@garrettmoon this would be also nice to check on CI. Thanks!

PR #600, opened 2021-05-28
Decode WebP images directly into a buffer Core Animation likes

Adlai-Holler:
Without this change, the images we output aren't compatible with the GPU and so CA has to copy them all on the main thread at commit time, which means stutters. Plus memory bloat.

The difference here is so substantial that you may seriously want to consider re-running a WebP experiment if you did in the past and bailed on it.

`CA::copy_image` before: 
Screen Shot 2021-05-28 at 5 11 00 PM

After:
Screen Shot 2021-05-28 at 5 11 24 PM

~Yes I am too lazy to figure out how to do this on macOS, and for that I am sorry. Desktops are fast anyway!~

### DIFF
--- a/Source/Classes/Categories/PINImage+DecodedImage.m
+++ b/Source/Classes/Categories/PINImage+DecodedImage.m
@@ -122,12 +122,6 @@ NSData * __nullable PINImagePNGRepresentation(PINImage * __nonnull image) {
         return nil;
     }
     
-#if PIN_WEBP
-    if ([data pin_isWebP]) {
-        return [PINImage pin_imageWithWebPData:data];
-    }
-#endif
-    
     PINImage *decodedImage = nil;
     
     CGImageSourceRef imageSourceRef = CGImageSourceCreateWithData((CFDataRef)data, NULL);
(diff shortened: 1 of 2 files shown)


Adlai-Holler:
Also updated the PR to attempt to use CGImageSource – which supports WebP at least as of iOS 14 – _before_ trying our own manual WebP decoding.

Adlai-Holler:
Looks like the CI failure is a config issue unrelated to the CL.

WDYT about the change @garrettmoon ? Also hi! Long time no code.

garrettmoon:
Long time no code!

I think the CI issue is real?
`    - WARN  | [PINRemoteImage/WebP] xcodebuild:  /Users/runner/work/PINRemoteImage/PINRemoteImage/Source/Classes/Categories/PINImage+WebP.m:15:13: warning: unused function 'releaseData' [-Wunused-function]`

I think the releaseData method needs to be ifdef'd out for iOS?

My only concern about this PR is the bit rot that will happen on macOS… Any chance you can try and unify some of the code for macOS in a follow up?

Adlai-Holler:
OK updated – lmk

garrettmoon (approved):
Looks good to me, thank you @Adlai-Holler !

You… uh… think you can apply this to the animated webp code path next? :D

PR #613, opened 2021-11-24
Add support for dynamic placeholder images

bdolman:
Before this change, if you attempted to use a dynamic image for a placeholder
(e.g. an image from an asset catalog that has both dark and light versions)
you would get unpredictable results. This is because PINAnimatedImageView
is drawing the image directly to the layer without consulting the view's
current trait collection. So which version of the image you get depends
entirely on what the global `UITraitCollection.currentTraitCollection` is.

According to Apple, "UIKit updates the value of this property before calling
several well-known methods of UIView, UIViewController, and UIPresentationController."
Though they do not say what those methods are, experimentation concludes
that "displayLayer:" is not one of those.

To solve this, we need to make sure that when we generate the CGImage,
we are explicit about using the view's trait collection so that we
get the correct image.


hovox:
hey @garrettmoon , is PINRemoteImage is still relevant to you and Pinterest? You are very passive here and it makes a felling that project is dead or almost dead. Maybe you can transition the project to someone else in that case or please let us know what you think about it ? We use it and all PRs and suggestions hanging and waiting months or years to your response.

Sorry if I sound offending, I just want to understand can we make project to move faster. I know we can make fork and all that, just let us know what you think.

garrettmoon:
Looks great, thank you for the contribution!

It sounds like this is a non-issue for the UIImageView categories?

PR #616, opened 2022-01-27
Update MD5 generation to CryptoKit Insecure.MD5

RRUIZ5:
When setting minimum deployment target to iOS >= 13, a build error is generated about CC_MD5 being deprecated. This PR updates the usages to CryptoKit's `Insecure.MD5`.

The functionality is the same, but through the usage of a newer API, now the framework builds for iOS >= 13


Source/Classes/Categories/NSString+MD5.swift
```diff
@@ -0,0 +1,33 @@
...
+//
+
+import CryptoKit
+import Foundation
+
+extension NSString {
+
+    private func MD5(with format: String) -> NSString {
+        guard #available(iOS 13, *),
+              #available(tvOSApplicationExtension 13.0, *) else { return "" }
```
garrettmoon, line 16: Should we be throwing fatal errors instead of returning empty strings?

Tests/PINRemoteImage-tvOSTests-Bridging-Header.h
```diff
@@ -0,0 +1,4 @@
+//
+//  Use this file to import your target's public headers that you would like to expose to Swift.
```
garrettmoon, line 2: Is this file necessary?

Tests/NSString+MD5Tests.swift
```diff
@@ -0,0 +1,40 @@
...
+class NSString_MD5Tests: XCTestCase {
+
+    override func setUpWithError() throws {
+    }
+
+    override func tearDownWithError() throws {
+    }
+
+    func testCryptoKitCacheKeyMD5() throws {
+        let testString: NSString = "My awesome test string"
```
garrettmoon, line 21: One additional test that might be useful is ensuring that cryptoKitCacheKeyMD5 returns the same hash as hashCacheKey_removeMeIOS13?

Tests/PINRemoteImage-tvOSTests-Bridging-Header.h
RRUIZ5: It was auto created by Xcode when I added the Swift test, should I try to remove it?

garrettmoon: I guess if it still is working without it 😅

Source/Classes/Categories/NSString+MD5.swift
RRUIZ5: done!
I also found a new way to generate data object from string that avoids optional, so one less check is needed

Tests/PINRemoteImage-tvOSTests-Bridging-Header.h
RRUIZ5: It turns out it is needed for the tests to run on tvOS target

Tests/NSString+MD5Tests.swift
RRUIZ5: done!

PINRemoteImage.xcodeproj/project.pbxproj
```diff
@@ -1105,6 +1143,7 @@
 		68A0FC221E523434000B552D /* Release */ = {
 			isa = XCBuildConfiguration;
 			buildSettings = {
+				ALWAYS_EMBED_SWIFT_STANDARD_LIBRARIES = YES;
```
garrettmoon, line 1146: Are we sure we want this to be YES?

Source/Classes/PINRemoteImageManager.m
```diff
@@ -1556,6 +1557,32 @@ - (NSString *)cacheKeyForURL:(NSURL *)url processorKey:(NSString *)processorKey
     return [self cacheKeyForURL:url processorKey:processorKey resume:NO];
 }
 
+-(NSString *) hashCacheKey_removeMeIOS13:(NSString *) string {
```
garrettmoon, line 1560: nit: to match code style elsewhere, add a space after the - and remove the ones after the )

Tests/PINRemoteImageManagerTests.m
```diff
@@ -0,0 +1,54 @@
...
+
+- (void)tearDown
+{
+}
+
+- (void)testCCMD5AndCryptoKitCacheKeyGenerateSameKey
+{
+    PINRemoteImageManager *manager = [PINRemoteImageManager sharedImageManager];
+
+    int num = 250;
```
garrettmoon, line 35: Certainly thorough 😅

PINRemoteImage.xcodeproj/project.pbxproj
RRUIZ5: No 😅
All of those are changes automatically done by Xcode
Do you want me to revert them and leave only those that introduced the new files?

RRUIZ5: Offline discussion took place, we decided to leave all and unset `ALWAYS_EMBED_SWIFT_STANDARD_LIBRARIES`

Source/Classes/PINRemoteImageManager.m
RRUIZ5: done!

PINRemoteImage.podspec
```diff
@@ -24,21 +24,21 @@ Pod::Spec.new do |s|
...
+
   ### Subspecs
   s.subspec 'Core' do |cs|
     cs.dependency 'PINOperation'
     cs.ios.deployment_target = ios_deployment
     cs.tvos.deployment_target = tvos_deployment
     cs.osx.deployment_target = osx_deployment
-    cs.source_files = 'Source/Classes/**/*.{h,m}'
-    cs.public_header_files = 'Source/Classes/**/*.h'
+    cs.source_files = 'Source/*/**/*.{swift,h,m}'
```
garrettmoon, line 36: I think the first `*/` can be removed here since `**` will include all directories?

PINRemoteImage.podspec
```diff
@@ -24,21 +24,21 @@ Pod::Spec.new do |s|
...
+    cs.public_header_files = 'Source/*/**/*.h'
```
garrettmoon, line 37: Same here, I think this can be `'Source/**/*.h'`

PINRemoteImage.xcodeproj/project.pbxproj
```diff
@@ -964,6 +1009,8 @@
 				PROVISIONING_PROFILE_SPECIFIER = "";
 				SDKROOT = appletvos;
 				SKIP_INSTALL = YES;
+				SWIFT_OPTIMIZATION_LEVEL = "-Onone";
```
garrettmoon, line 1012: Is this only for the debug build? We don't want this to be set for release versions I don't think.

PINRemoteImage.xcodeproj/project.pbxproj
```diff
@@ -1040,6 +1090,9 @@
 				PRODUCT_NAME = "$(TARGET_NAME)";
 				PROVISIONING_PROFILE_SPECIFIER = "";
 				SDKROOT = appletvos;
+				SWIFT_OBJC_BRIDGING_HEADER = "Tests/PINRemoteImage-tvOSTests-Bridging-Header.h";
+				SWIFT_OPTIMIZATION_LEVEL = "-Onone";
```
garrettmoon, line 1094: Same question here.

PINRemoteImage.xcodeproj/project.pbxproj
```diff
@@ -1279,13 +1336,16 @@
 				MODULEMAP_FILE = Source/PINRemoteImage.modulemap;
 				PRODUCT_BUNDLE_IDENTIFIER = com.pinterest.PINRemoteImage;
 				SKIP_INSTALL = YES;
+				SWIFT_OPTIMIZATION_LEVEL = "-Onone";
```
garrettmoon, line 1339: And here 😅

Tests/PINRemoteImageTests.m
```diff
@@ -392,12 +398,14 @@ - (void)testImageRequestIgnoresLocalData
...
-    [self.imageManager setRequestConfiguration:^NSURLRequest * _Nonnull(NSURLRequest * _Nonnull request) {
+
+    // To help reduce possibilities of flakiness, assign directly
+    // instead of `setRequestConfiguration` to avoid locks
+    self.imageManager.requestConfigurationHandler = ^NSURLRequest * _Nonnull(NSURLRequest * _Nonnull request) {
         XCTAssertEqual(request.cachePolicy, NSURLRequestReloadIgnoringLocalCacheData);
         [expectation fulfill];
         return request;
-    }];
+    };
```
garrettmoon, line 409: I think if we avoid using the locks we may create flakiness in a different way :( Does calling `[self waitForExpectationsWithTimeout:[self timeoutTimeInterval] handler:nil];` after calling setRequestConfiguration: solve the flakiness?

garrettmoon: Oh man, I'm so sorry this is designed this way 😅 

I think we need to add a new API:
`setRequestConfiguration:completion` (which the old `setRequestConfiguration:` can call with a nil completion). Then we can call the new API in the test method with a new expectation that is fulfilled in the completion block. We'll then wait for just that expectation to be fulfilled before calling `downloadImageWithURL:…`

RRUIZ5:
Closing since we decided not to continue with MD5, extracted the stability changes unto #618

PR #617, opened 2022-02-02
Revert "Disable asserts in release builds when using Swift Package Manager"

mohamed3114:
Reverts pinterest/PINRemoteImage#615


garrettmoon:
Can you say why this is a good idea? Generally asserts should be disabled in release builds.

PR #641, opened 2024-05-15
Support Building from Xcode 15

andyfinnell:
## Summary

Currently, PINRemoteImage does not build with recent build tools. We should fix that, plus update all the dependencies.

This bumps min versions of iOS and tvOS up to 14.0, and macOS to 11.0. This allows us to drop the `libwebp` dependency, which is not currently building with recent tools either, and instead use the system provided decoder. **This does mean tvOS loses support for WebP as tvOS doesn't have a system decoder.**

Also, clean up the headers. Cocoapods was including way too much as public headers, and we were using double quotes instead of angle brackets for them.

## Test plan

Make sure everything builds & unit tests pass. Also, update all the examples and ensure they work.


andyfinnell:
cc @tstump-phunware

jaichaudhry:
Is this planned to be in any release soon?

andyfinnell:
Hi @jaichaudhry . Yes, this got released with [3.0.4](https://github.com/pinterest/PINRemoteImage/releases/tag/3.0.4).

PR #656, opened 2026-07-29
Serialize animated-image frame decodes; refuse damaged frames; fix fr…

dreed1:
…ame-failure bookkeeping

CGImageSource is not safe for concurrent lazy decodes of the same source (despite the optimistic '*seems* immutable' comments), and PINCachedAnimatedImage reaches imageAtIndex:cacheProvider: concurrently from its caching queue, its init-time warmup dispatch, and coverImage callers. The racing decode crashes inside ImageIO
(GIFReadPlugin::copyImageBlockSetImp / GIFBufferInfo memmove).

- GIF, APNG, and WebP now serialize all ImageIO calls on the shared source — including CGImageSourceGetStatusAtIndex, which advances parser state — behind a per-image lock.
- Refuse affirmatively damaged frames (kCGImageStatusInvalidData / kCGImageStatusUnexpectedEOF) before attempting creation. Deliberately narrow: kCGImageStatusIncomplete (trailer-less but renderable GIFs, common in the wild) still decodes.
- A frame that fails to decode previously left _frameRenderCount held and its index parked in _cachedOrCachingFrames forever: playbackReady never fired and the frame was never retried. Failures now release their render slot.
- pin_decodedImageRefWithCGImageRef: returned the borrowed input ref when CGBitmapContextCreate fails (memory pressure); animated callers then released it and returned the dangling pointer. The fallback now matches the success path's +0 autoreleased contract. Also guard NULL input.

Fixes a long-standing top crasher in a large production app (EXC_BAD_ACCESS at PINImage+DecodedImage's CGContextDrawImage during GIF frame decode).

### DIFF
--- a/Source/Classes/AnimatedImages/PINAPNGAnimatedImage.m
+++ b/Source/Classes/AnimatedImages/PINAPNGAnimatedImage.m
@@ -31,6 +31,7 @@
     size_t _loopCount;
     CFTimeInterval *_durations;
     NSError *_error;
+    NSLock *_decodeLock; // serializes frame decodes on _imageSource
 }
 @end
 
@@ -40,6 +41,7 @@
 {
     if (self = [super init]) {
         _animatedImageData = animatedImageData;
+        _decodeLock = [[NSLock alloc] init];
         _imageSource =
             CGImageSourceCreateWithData((CFDataRef)animatedImageData,
                                         (CFDictionaryRef)@{(__bridge NSString *)kCGImageSourceTypeIdentifierHint:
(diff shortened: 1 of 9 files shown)


Source/Classes/AnimatedImages/PINAPNGAnimatedImage.m
```diff
@@ -31,6 +31,7 @@ @interface PINAPNGAnimatedImage ()
     size_t _loopCount;
     CFTimeInterval *_durations;
     NSError *_error;
+    NSLock *_decodeLock; // serializes frame decodes on _imageSource
```
andyfinnell: It looks like this project uses `PINRemoteLock` for locks.

Source/Classes/Categories/PINImage+DecodedImage.m
```diff
@@ -253,9 +259,15 @@ + (CGImageRef)pin_decodedImageRefWithCGImageRef:(CGImageRef)imageRef
...
+
     }
-    
+
+    // match the success path's +0 autoreleased contract.
+    // Returning the borrowed imageRef directly made the animated-image callers release
+    // it and then return the same (now dangling) pointer whenever CGBitmapContextCreate
+    // failed under memory pressure — an over-release/use-after-free.
+    CGImageRetain(imageRef);
+    CFAutorelease(imageRef);
```
andyfinnell, line 270: Aren't these no-ops these days? It's been a while since I've done CoreFoundation level stuff.

andyfinnell (approved):

Source/Classes/Categories/PINImage+DecodedImage.m
dreed1: I had to look this up, cuz it's a good point --  ARC only manages Objective-C object pointers. CGImageRef is a CoreFoundation-style type, and ARC never touches CF references that aren't explicitly bridged. I guess we still have to manually count references to stuff like this, i wonder if there's more of these littering around our code

Source/Classes/AnimatedImages/PINAPNGAnimatedImage.m
dreed1: good point, thank you!

andyfinnell (approved):

Source/Classes/Categories/PINImage+DecodedImage.m
andyfinnell: Ah, that's a good learning.

